PR #2, opened 2015-04-29
keypad support on login page

conejoninja:
Added keypad support, works in the same way as in the webwallet, (ie. 9 represent the top right number of the screen....)

### DIFF
--- a/login.html
+++ b/login.html
@@ -291,7 +291,7 @@ function pinAdd(num)
 	e.value += num;
 }
 
-function pinBackspace(num)
+function pinBackspace()
 {
 	var e = document.getElementById('pin');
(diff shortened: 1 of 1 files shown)


login.html
```diff
@@ -441,6 +441,31 @@
...
 
+
+function pinKeydownHandler(e) {
+    var k = e.keyCode;
+    if (k === 8) { // Backspace
+        pinBackspace();
+    } else if (k === 13) {
+        pinEnter();
+    } else if (isNumericKey(k)) {
+        num = getNumberFromKey(k);
```
jpochyla: Missing `var` here?

conejoninja: Yes, fixed now. Also removed some debugging lines I forgot to.

PR #4, opened 2015-05-13
adding CSS styles

runn1ng:
I have changed styles from inline style to explicit CSS rules, so it will be easier for other people to restyle the look


prusnak:
Thx!

PR #5, opened 2015-08-16
Version method 2

conejoninja:
Get the correct API version from the firmware version


prusnak:
I don't like the change. It puts lots of extra requirements on the version. For example it assumes there will be no version such as 1.11.37.

prusnak:
Implemented semantic version compare in https://github.com/trezor/connect/commit/e14dba5f53ae8d17421f107d9ea8b36c95886087

conejoninja:
much better :+1:

PR #13, opened 2015-12-17
Add server side C# implementation of Trezor Connect

NicolasDorier:
Add server side sample C# implementation of Trezor Connect


prusnak:
Thanks!

PR #14, opened 2016-01-28
Releasing

runn1ng:
Releasing the Trezor right after action


popup/popup.js
```diff
@@ -137,12 +137,14 @@ function handleLogin(event) {
             let {message} = result;
             let {public_key, signature} = message;
 
-            respondToEvent(event, {
-                success: true,
-                public_key: public_key.toLowerCase(),
-                signature: signature.toLowerCase(),
-                version: 2      // since firmware 1.3.4
-            });
+            return device.session.release().then(() => {
```
runn1ng: I am using return here so if the release failed, I still fall through to the catch block afterwards

PR #49, opened 2017-03-23
add zcash xpub export

xbach:


runn1ng:
Seems great, luckily, zcash has the same xpub magic :)

Issue #11, opened 2015-12-01
Support  'Show on TREZOR' 


matiu:
are requested in https://github.com/bitpay/copay/issues/3558

zachoooo:
Also requested in kvhnuke/etherwallet#651

prusnak:
@szymonlesisz is already working on it

szymonlesisz:
already done in v3

PR #53, opened 2017-07-03
Feature/get address

szymonlesisz:


runn1ng:
Fixes https://github.com/trezor/connect/issues/11

runn1ng:
Hm, the segwit checkbox in the example doesn't do anything and still shows me non-segwit address

runn1ng:
I am not sure where is the error, since the code looks fine

runn1ng:
Hm, seems like connect is using old trezor.js. Will try to update and rebuild

PR #56, opened 2017-07-09
Add new API `setCoinName()`.

cryptcoin-junkey:
Altcoins can't use connect.js since some API uses always `Bitcoin` as the coin name. 
IMO, the best way to fix this issue is implementing API version 3. This patch is a hotfix.

This patch is not tested well. Forgive me if there have some minor bugs.


prusnak:
This is not really nice way of handling things.

NACK from me.

We need to handle this more nicely by adding coin parameter to API calls and not introduce state.

cryptcoin-junkey:
Well. I also think the best way is to re-define API (possibly v3). But, as a practical story, it must be a hard work. And API compatibility breakage (like v2.1) may causes some side effects to current users.

Could you tell me your road map around this repository?

prusnak:
There can't be any compatibility breakage, so this should become a part of v3 API spec.

There is no fixed roadmap, but we are in the process of redesigning the API. Being able to specify the coin would be one of the main points.

Would you be interested in designing the new v3 API? We planned to do this internally, but I guess we can do this in the open if enough external people are interested.

cryptcoin-junkey:
My interest is to implement a online wallet service that supports Trezor rather than API development.
But I can contribute something from the viewpoint of API user if it was opened as alpha, beta, or something pre-release.

For now, I see this fix is unacceptable and will be fixed by v3 API. I close this. I’m looking forward to the next release.

runn1ng:
@cryptcoin-junkey Trezor Connect was originally meant just for secure login and it kept growing organically, which created a little strange API

We are now creating a new Connect that will be easier to maintain, will have easier APIs and will allow easier integration of new coins. I will let you know when this will be done.

PR #59, opened 2017-07-31
Claimbch

szymonlesisz:


examples/claim-bcc.html
```diff
@@ -0,0 +1,67 @@
...
+                return s;
+            }
+        }).join('/');
+    }
+
+    function calculateFee(unspents){
+
+    }
+
+    window.TREZOR_POPUP_URL = 'http://localhost/popup/popup.html';
```
runn1ng: this should probably not be in examples

runn1ng: examples can be run from connect.trezor.io... so as in the other examples this shouldn't be here

not a blocker, but would be nice to fix

popup/popup.js
```diff
@@ -1638,6 +1768,35 @@ function waitForAccount() {
...
+    };
+
+    let onUsed = () => {};
+
+    let onEnd = () => {
+        discovered.push(discovering);
+        discovering = null;
+    };
+
+    return new Promise((resolve, reject) => {
```
runn1ng: nitpick

you don't need to create new Promise, you can just do

````
return discoverAccounts(...).then(() => discovered)
````

runn1ng: Also I am not sure what this promise does... it returns just the first account? Why do you want that?

runn1ng: Why not just return `discoverAccounts(...)`? 

Again not blocker, since the claiming code doesn't use that at all

PR #64, opened 2017-08-01
Fix referencedtx

runn1ng:


runn1ng:
Apparently this produces some errors. Need to find, fix.

PR #69, opened 2017-08-09
sign and verify ethereum message

szymonlesisz:


fabioberger:
@runn1ng is a new version going to be available at:

```
<script src="https://connect.trezor.io/4/connect.js"></script>
```

?

runn1ng:
Most likely it will be at `/3/`, since we can change `/3/` because nobody is using it now except for us (in Bitcash claim tool).

fabioberger:
@runn1ng @szymonlesisz am I really able to test this change locally? In `connect.js`, there are these lines:

```
var POPUP_URL = window.TREZOR_POPUP_URL || 'https://connect.trezor.io/' + VERSION + '/popup/popup.html';
    var POPUP_PATH = window.TREZOR_POPUP_PATH || 'https://connect.trezor.io/' + VERSION;
    var POPUP_ORIGIN = window.TREZOR_POPUP_ORIGIN || 'https://connect.trezor.io';
```

Must I set these variables to get it to work?

szymonlesisz:
you need to set a localhost server to the root path of connect
then in the html file you need to set:

<script>
        window.TREZOR_POPUP_PATH = 'https://localhost/';
        window.TREZOR_POPUP_ORIGIN = 'http://localhost';
        window.TREZOR_POPUP_URL = 'http://localhost/popup/popup.html';
</script>
<script src="http://localhost/connect.js"></script>

PR #74, opened 2017-09-03
NEM Support

saleemrashid:
See trezor/trezor-mcu#196

### DIFF
--- a/connect.js
+++ b/connect.js
@@ -542,6 +542,25 @@ this.TrezorConnect = (function () {
             }, requiredFirmware), callback);
         }
 
+        this.nemSignTx = function (
+          address_n,
+          transaction,
+          callback,
+          requiredFirmware
+        ) {
+            if (requiredFirmware == null) {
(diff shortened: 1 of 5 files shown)


prusnak:
Yes, please, merge this before ...

szymonlesisz:
Hi @saleemrashid, is there any reason why you put "show_display" parameter into "nemGetAddress" method? 
Im asking because in other "getAddress" methods we always prompt this message before user is asked for PIN and its mandatory, but in your code is optional.
I've just wrote some examples and it returns proper results (same results as in python-trezor tests)

saleemrashid:
@szymonlesisz I think it was because `nemGetAddress` is more similar to `getXpubKey` than `getAddress` (because NEM addresses are reused). But in hindsight, I think we should enforce `show_display` and remove the optional parameter from `nemGetAddress`.

saleemrashid:
Apparently this was merged ad427bf895ca2361dc57fbee8c7add20a37b64e3

PR #80, opened 2017-10-01
popup: Support file:// URLs

saleemrashid:
Allows use of TREZOR Connect from file:/// URLs. Tested in Firefox and Chromium with MyEtherWallet v3.10.4.

/cc @prusnak

Firefox
Chromium

### DIFF
--- a/popup/popup.js
+++ b/popup/popup.js
@@ -164,6 +164,11 @@ const CHROME_EXTENSION_NAMES = {
 };
 
 function parseIdentity(event) {
+    // file:// URLs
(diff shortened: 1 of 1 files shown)


runn1ng:
I vaguely remember I solved something like this before

Oh, I know, in trezord, which does similar thing. electron apps, chrome extensions, and local apps all put `null` as origin of http request.

So I *think* the `chrome-extension` below is useless anyway, since crome extensions return null. (They didn't use to, but Chrome changed it.)

PR #90, opened 2017-11-17
Show NEM address on device

saleemrashid:
This was what I intended in https://github.com/trezor/connect/pull/74#issuecomment-336456040


prusnak:
Thanks!

PR #91, opened 2017-11-17
Do not hexlify NEM messages

saleemrashid:
TREZOR Connect expects NEM transactions in NIS format where the payload is already hex-encoded.


prusnak:
Thanks!

PR #97, opened 2017-12-16
Add NEM message decryption

saleemrashid:


connect.js
```diff
@@ -587,6 +587,29 @@ this.TrezorConnect = (function () {
...
+        this.nemDecryptMessage = function (
+          address_n,
+          network,
+          public_key,
+          payload,
+          callback,
+          requiredFirmware
+        ) {
+            if (requiredFirmware == null) {
+                requiredFirmware = '1.6.1'; // first firmware that supports NEM message decryption
```
saleemrashid: @prusnak Will this be correct?

prusnak: Yes

szymonlesisz:
@saleemrashid could you provide some example for this call? 
I was trying to find it in python-trezor tests but there is none. I need some values to "public_key" and "payload" fields to test it with connect.
I assume that "public_key" will be xpub for path given in first parameter. Is that right?
Wouldn't it be better to let connect find this xpub from given path automatically? This way we can avoid the need to double call (first to get xpub, second to sign message)?

So it would be look like this:
TrezorConnect.nemDecryptMessage(path, network, payload, callback);

Thoughts?

saleemrashid:
It is the message sender's public key. The firmware combines this with the private key to derive a shared secret and decrypt the message.

https://github.com/trezor/trezor-common/blob/c88d7b962a8494403ca562c2f3373a917c370bdb/protob/messages.proto#L887

szymonlesisz:
@saleemrashid Great, so i will leave it as it is. What about the first part of my response?
Could you give me some values from ALL-ALL-ALL device so i will be able to make an example in html?

PR #99, opened 2017-12-31
Add prefix to global VERSION variable

fabrik42:
Hi there!

First of all, thank you for keeping all your software open and enabling others to tinker with it!

While getting my feet wet with this lib, I realised that the `VERSION` variable is leaking into the global namespace. It is a very generic name so this could lead to problems with other libraries.

This is a first small fix that adds a prefix to the variable, I tried to do this according to the naming of the other global vars.

I think another solution could be to move the variable into the scope of [window.TrezorConnect](https://github.com/trezor/connect/blob/master/connect.js#L24) as it does not seem to be used anywhere else. This would mean that the version and the URL of the popup can not be determined from the outside. I am not sure if this is a valid use case: to load connect.js v4 but require the popup of another version?

Keep up the good work!

Best,
Christian

### DIFF
--- a/connect.js
+++ b/connect.js
@@ -3,7 +3,7 @@
  *
  * GPLv3
  */
-var VERSION = 4;
+var TREZOR_CONNECT_VERSION = 4;
 
 if (!Array.isArray) {
     Array.isArray = function(arg) {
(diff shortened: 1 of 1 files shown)


szymonlesisz:
Merged 
https://github.com/trezor/connect/commit/bcf7e2386f46874bcb836699343f2e6f66e7b5cf
and deployed to https://connect.trezor.io/4/

Thanks!

PR #103, opened 2018-02-03
Update xpubkey-etherium example

eric-smyth:
The original path provided will not return the correct values from the trezor, this fixes the path.


prusnak:
It will return the correct xpub. But then you need to do m/0/0 to get
the address, instead of m/0 what it seems you are doing.

eric-smyth:
Bah, you are right, sorry - I'm just trying to ramp up on this stuff.

PR #105, opened 2018-02-20
Adding support for Flashcoin

flashbountyhunter:


prusnak:
This file is not supposed to be edited manually. Wait until your PR to trezor-common is merged.

PR #123, opened 2018-05-23
Change URL for Bitcore (Monacoin).

cryptcoin-junkey:
mona.chainsight.info is not maintained well and it cannot handle SegWit transactions.
Newer URL can handle them.

### DIFF
--- a/popup/coins.json
+++ b/popup/coins.json
@@ -430,7 +430,7 @@
 	"min_address_length": 27,
 	"max_address_length": 34,
 	"bitcore": [
-		"https://mona.chainsight.info"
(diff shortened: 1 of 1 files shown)


prusnak:
The correct way is to send changes to these two files:

https://github.com/trezor/trezor-common/blob/master/coins.json
https://github.com/trezor/trezor-common/blob/master/defs/coins/monacoin.json

prusnak:
Done in https://github.com/trezor/trezor-common/pull/111

cryptcoin-junkey:
@prusnak Thanks for your comment. I'm going to move to trezor/trezor-common#111.

PR #147, opened 2018-06-26
Add Viacoin to coins.json

romanornr:


romanornr:
Oh reading this now: https://github.com/trezor/connect/pull/105
Seems like it has to be done via Trezor common

prusnak:
Viacoin is already in trezor-common.

PR #163, opened 2018-07-27
Add 'isHex' function 

mlejva:
for testing whether a string is in a hexadecimal format


szymonlesisz:
it's already there:
https://github.com/trezor/connect/blob/develop/src/js/utils/ethereumUtils.js#L40

PR #184, opened 2018-08-11
Tron Integration

Rovak:
This PR adds the required methods to support the Tron Protocol

It works but still needs some code cleanup like adding types and removing some ethereum specific code that was copied. I would like some feedback if everything is in the right place and if i missed anything.


szymonlesisz:
Hi @Rovak,
Thanks! Very nice PR!

Waiting for https://github.com/trezor/trezor-common/pull/166

prusnak:
There is no code for Tron in the firmware, so it does not make sense to merge this.

PR #195, opened 2018-08-18
Changes for Wanchain

tyrion70:


tyrion70:
This pull request completes the other pull requests already made for Wanchain. The extra field txtype could not be added before trezor-js was deprecated because v4 was still depending on that library. with v5 the change is very easy.

szymonlesisz:
looks legit, thanks!

PR #214, opened 2018-09-13
Tezos: Added tests & removed curve property

jurajselep:
I will add more tests & documentation


szymonlesisz:
First of all i shouldn't accept this PR at all because it was way behind develop branch and it has multiple conflicts and bugs.
Yet i had a good day today and i gave it some time to fix some major issues for you.
Here's my commits:
https://github.com/trezor/connect/commits/feature/tezos-fixed

Unfortunately that's not all.
There are still missing documentation and test for TezosSignTransaction operations others than "transaction".

There is also few flowtype fields that are unclear. For example in ./types/tezos.js: $TezosSignTransaction you had  declared "branch" field as Array<number> and in test you are using a string.

Please validate ALL Tezos types which are placed at ./types/trezor.js now.

jurajselep:
thanks, I will fix issues and create new pull request

PR #215, opened 2018-09-14
ethereumSignMessage - Skip hex conversion if message is already hex

cavanmflynn:
To prevent the data loss issue explained in https://github.com/trezor/connect/issues/216, Trezor Connect should check if the data passed into `ethereumSignMessage` is already hex and only convert if necessary.


charlielye:
Thanks for this. This blocks a whole range of useful use cases (e.g. feeless transactions).
@prusnak ETA on merging?

szymonlesisz:
It's going to be merged today and released probably on wednesday (tomorrow)

szymonlesisz:
I've added the same way to ethereumVerifyMessage, otherwise you will not be able to verify signed message. Also i've fixed etherumUtils to allow '0x' prefix with uppercase '0X'.
Closing this in favour of https://github.com/trezor/connect/pull/219

PR #229, opened 2018-10-05
Removed 2 hidden unicode characters in regex

### LINKED ISSUE #227: Connect popup goes away after a few seconds (Iframe timeout)
**This might be limited to minified builds.** I'm using Trezor Connect v5.0.33 with default react-create-app build scripts. Specifically I am calling `trezor.ethereumGetAddress`. Not completely sure what is going on here but:
 - When accessing a served dev build everything works fine
 - When accessing a served production build the window opens, displays "Loading" for a few seconds and then closes with a `Iframe timeout` error. I also get an error as soon as the window opens saying `Failed to execute 'postMessage' on 'Window': Invalid target origin 'undefined' in a call to 'postMessage'.`

I worked around this by including the trezor-connect.js version from CDN instead of importing the NPM module. Not sure what is going on but I figured this might be helpful to someone.

mohammadrafigh:
Fixes #227 
these characters will turn into unicode format in production builds, something like these:
/^.+\:\/\/[^\\u200c\u200b/]+/

so the library won't work in production builds with an iframe timeout.

### DIFF
--- a/src/js/iframe/builder.js
+++ b/src/js/iframe/builder.js
@@ -42,7 +42,7 @@ export const init = async (settings: ConnectSettings): Promise<void> => {
     }
 
     // eslint-disable-next-line no-irregular-whitespace, no-useless-escape
-    const iframeSrcHost: ?Array<string> = instance.src.match(/^.+\:\/\/[^\‌​/]+/);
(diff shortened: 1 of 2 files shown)


prusnak:
Thank you! Merged to `develop` and cherry-picked to `v5`.

PR #254, opened 2018-11-06
Add Eos integration

atkachyshyn:


prusnak:
Depends on https://github.com/trezor/trezor-common/pull/231

atkachyshyn:
Hey guys, can you send us instructions on how to install connect on your dev server, using "make build-test"? It seems we need to pass some authorization but it is not said in the instructions.

atkachyshyn:
@prusnak Hey Pavol. Any plans on having the pull request reviewed any time soon? Can you guys update us on this matter, please?

szymonlesisz:
i'm on it right now. 
Since this PR was a little bit old i've made some changes in this branch:
https://github.com/trezor/connect/tree/pr/eos

Currently i'm reviewing `eosSignTrnasaction` method and i've found that there is a couple of differences between field names in `trezor-connect` parameters and [official Eos javascript sdk](https://eosio.github.io/eosjs/) 

for example in `connect` you've defined
```javascript
type EosActionDelegate = {
    sender: number,
    receiver: number,
    netQuantity: EosAsset,
    cpuQuantity: EosAsset,
    transfer: boolean,
}
```

but in `eosjs` it looks like: (note the snake case and different field names)
```javascript
{
    account: 'eosio',
    name: 'delegatebw',
    authorization: [{
      actor: 'useraaaaaaaa',
      permission: 'active',
    }],
    data: {
      from: 'useraaaaaaaa',
      receiver: 'mynewaccount',
      stake_net_quantity: '1.0000 SYS',
      stake_cpu_quantity: '1.0000 SYS',
      transfer: false,
}
```

Since those objects gets eventually translated inside [signTx helper](https://github.com/trezor/connect/blob/pr/eos/src/js/core/methods/helpers/eosSignTx.js#L124) into Trezor protobuf format, don't you want to have those two libraries compatible to avoid unnecessary translating back and forth?

It's not a blocker, i can review it as it is, i'm just curious

szymonlesisz:
FYI i'm working on `eosSignTransaction` params to be more transparent and compatible with eos api

atkachyshyn:
Hey Szymon, yes, it makes sense. Creating this pull request we were working in parallel on other pull requests and also web wallet, so we had to stick to some style, and camel case was just more natural and to be more distinguishable between trezor and eos contracts. It's okay if you are changing, we will just make according changes on our integration. It's great that you have started the review, thanks :) If something needs to be changed just leave the comment, and i'll pick this up ASAP. I won't change params since you are working on it right now, otherwise just leave the comments.

cryptosig:
> i'm on it right now.
> Since this PR was a little bit old i've made some changes in this branch:
> https://github.com/trezor/connect/tree/pr/eos
> 
> Currently i'm reviewing `eosSignTrnasaction` method and i've found that there is a couple of differences between field names in `trezor-connect` parameters and [official Eos javascript sdk](https://eosio.github.io/eosjs/)
> 
> for example in `connect` you've defined
> 
> ```js
> type EosActionDelegate = {
>     sender: number,
>     receiver: number,
>     netQuantity: EosAsset,
>     cpuQuantity: EosAsset,
>     transfer: boolean,
> }
> ```
> 
> but in `eosjs` it looks like: (note the snake case and different field names)
> 
> ```js
> {
>     account: 'eosio',
>     name: 'delegatebw',
>     authorization: [{
>       actor: 'useraaaaaaaa',
>       permission: 'active',
>     }],
>     data: {
>       from: 'useraaaaaaaa',
>       receiver: 'mynewaccount',
>       stake_net_quantity: '1.0000 SYS',
>       stake_cpu_quantity: '1.0000 SYS',
>       transfer: false,
> }
> ```
> 
> Since those objects gets eventually translated inside [signTx helper](https://github.com/trezor/connect/blob/pr/eos/src/js/core/methods/helpers/eosSignTx.js#L124) into Trezor protobuf format, don't you want to have those two libraries compatible to avoid unnecessary translating back and forth?
> 
> It's not a blocker, i can review it as it is, i'm just curious

I had heard/read that in order to combat signed transaction fraud they would/might be adding hashing of the transaction to make sure that transaction details were not changed in transit and unwittingly signing and broadcasting a transaction with unexpected results. Having a deviation from the eosjs now might be a problem in the near future.

szymonlesisz:
I did massive refactoring of `eosSignTransaction`. Here's what's changed:
- input data serialization from human readable format compatible with eosjs and python-trezor
- added tests, vectors are taken from [python-trezor](https://github.com/trezor/trezor-firmware/blob/master/python/trezorlib/tests/device_tests/test_msg_eos_signtx.py)
- rewritten `processTxRequest` - it's important to respond on device ack request correctly, host should not decide or remember data offset position on action index. Actually we've found that this is a [firmware integration issue](https://github.com/trezor/trezor-firmware/issues/247) and it will be probably changed too.


Please have a look on [this branch](https://github.com/trezor/connect/tree/pr/eos) and check if all parameters are correct and compatible. Feel free to update them if not

szymonlesisz:
@cryptosig 
> I had heard/read that in order to combat signed transaction fraud they would/might be adding hashing of the transaction to make sure that transaction details were not changed in transit and unwittingly signing and broadcasting a transaction with unexpected results. Having a deviation from the eosjs now might be a problem in the near future.


Yes but it's not only about compatibility with other lib, it's also about clarity of those params.
After my first look on this PR i didn't event know how to serialize parameters data and add them to tests. This was my intention to parameters be in more "human readable" form

szymonlesisz:
another thing that i want to ask you after your review is to add docs, especially for `eosSignTransaction`. I don't know Eos network very well

atkachyshyn:
Sorry for the delay. We have started looking to your changes, and started with running refactored connect locally, to verify it on EOS network. But we have got troubles connecting web wallet to local instance of connect, as it seems configuration via TrezorConnect.init() is not working properly. As I see you guys still have ongoing development\refactoring, so probably it will be fixed soon, if it's not already. But for us it ignores\overrides connectSrc property, and uses hosted version (connect.trezor.io/7/*), which therefore retrieves configs which do not know about existence of EOS and we cannot communicate with device. But we already bypass this, so we will get back to you soon. And of course we will add docs as you have requested.

atkachyshyn:
Changes are ok with us. We have intentionally didn't include conversions to types that Eos ecosystem expects, as it seemed not to be the right place as there is no similar logic for other currencies. I have adapted our integration according to your changes and signed a few transactions, and Eos network accepted signatures, so it seems everything is ok. 

Also i have added simple docs with descriptions, similar to other currencies. Do you need more details on each transactions or is it enough?

szymonlesisz:
closed via #412

PR #266, opened 2018-11-19
capricoin: propagating timestamp field

Adman:
Please, let me know if I missed something to add.
The related bitcoinjs-lib pull request is here https://github.com/karel-3d/bitcoinjs-lib/pull/10
I also opened PR in hd-wallet repo which can be found here https://github.com/trezor/hd-wallet/pull/29

I want to write a small test for this. Shall I write it with `all all...` mnemonic like in python-trezor or the one in connect's docs?

Signed-off-by: Adrian Matejov <a.matejov@centrum.sk>


knaperek:
Hi, any idea about when we could expect this to be merged? The core support for CPC is already released in the latest firmware and we also have a first web wallet integration ready, just waiting for this to be merged. Thanks.

szymonlesisz:
I'm on it. Could you provide some testing data for signTransaction method?
Including legacy and segwit inputs.
It will be great if you provide those data with our testing mnemonic:  "all all all all all all all all all all all all" so we can include this to automatic tests

tarpas:
Hi, @Adman is busy these days, he'll let us know if he has questions (what form would you like the data in?)

I have a general remark: when working on the firmware it was our first instinct to add many more tests for different types of transactions. However after feedback from stick and matejcik we deleted them (https://github.com/trezor/python-trezor/pull/325/commits). The change just adds one field to the transaction and doesn't change anything else. The existing code has enough coverage for all the cases it should handle. The timestamp field doesn't change anything in the execution flow. Segwit transactions and transactions without timestamp are not valid CPC transactions, so it would be beneficial if the library refused them. 

However the conclusion was clear and I think correct. Citing matejcik:

"actually we had an internal discussion and let's keep it as it is now, without the fail-test.

it should be there technically, but it is not worth complicating the trezor-core implementation over it"

So only one test got to the final commit. Is this one test also sufficient for you?

* https://github.com/testmon/python-trezor/blob/466e3878ea9bf7a93d3cf77b9b422d6503c95bb5/trezorlib/tests/device_tests/test_msg_signtx_capricoin.py
* https://github.com/testmon/python-trezor/blob/466e3878ea9bf7a93d3cf77b9b422d6503c95bb5/trezorlib/tests/txcache/insight_capricoin_tx_3bf506c81ce84eda891679ddc797d162c17c60b15d6c0ac23be5e31369e7235f.json
* https://github.com/testmon/python-trezor/blob/466e3878ea9bf7a93d3cf77b9b422d6503c95bb5/trezorlib/tests/txcache/insight_capricoin_tx_f3a6e6411f1b2dffd76d2729bae8e056f8f9ecf8996d3f428e75a6f23f2c5e8c.json

Adman:
@szymonlesisz I've added that test from python-trezor.
But I'm not sure where to specify all all... mnemonic. Can you point me to correct location?

When running this specifi test, I always get
```
SignTransactionCapricoin /signCPC FAILED
	Expected undefined to equal '01000000665ccf5b025f23e76913e3e53bc20a6c5db1607cc162d197c7dd791689da4ee81cc806f53b000000006b483045022100fce7ccbeb9524f36d118ebcfebcb133a05c236c4478e2051cfd5c9632920aee602206921b7be1a81f30cce3d8e7dba4597fc16a2761c42321c49d65eeacdfe3781250121021fcf98aee04939ec7df5762f426dc2d1db8026e3a73c3bbe44749dacfbb61230ffffffff8c5e2c3ff2a6758e423f6d99f8ecf9f856e0e8ba29276dd7ff2d1b1f41e6a6f3010000006a473044022015d967166fe9f89fbed8747328b1c4658aa1d7163e731c5fd5908feafe08e9a6022028af30801098418bd298cc60b143c52c48466f5791256721304b6eba4fdf0b3c0121021fcf98aee04939ec7df5762f426dc2d1db8026e3a73c3bbe44749dacfbb61230ffffffff01a0782d00000000001976a914818437acfd15780debd31f3fd21d4ca678bb36d188ac00000000'.
```

szymonlesisz:
this PR was a little bit outdated with develop, so i've rebase it and create new branch for it:
https://github.com/trezor/connect/tree/pr/capricoin

Also if was failing signtTransaction (bitcoin) test, since this line:
https://github.com/trezor/connect/blob/522063d7cbd6fe77efc86236833497302ef50322/src/js/backend/BlockBook.js#L179
was not written properly (in javascript `null` !== `undefined`)
This causes problems with other coins (hd-wallet:bitcore returns this field as undefined, not null)

Now your test should be working properly.

szymonlesisz:
We are almost there :)
do you have some more test cases? like:
- segwit (p2sh) transaction
- multiple outputs transaction

You can get inspired by:
https://github.com/trezor/connect/blob/develop/src/__tests__/core/signTransaction.spec.js

szymonlesisz:
Closing in favour of https://github.com/trezor/connect/tree/pr/capricoin

PR #272, opened 2018-11-30
Change network for protocol magic when signing tx on cardano

refi93:
We revisited the "network" attribute that was sent in the cardanoSignTransaction call and we suggest replacing it directly with the protocol magic, which makes it more flexible, since it allows you to sign a transaction for any cardano blockchain. This proves useful for custom cardano testnets.

Related to https://github.com/trezor/trezor-core/pull/417 and https://github.com/trezor/trezor-common/pull/244


prusnak:
Depends on https://github.com/trezor/trezor-common/pull/244

refi93:
updated trezor-common submodule to include related changes

refi93:
is there anything to fix/change so it can be merged?

szymonlesisz:
yes, sorry.
Since this change is not compatible with previous/current firmware it will be released together with the new firmware (2.0.11) as connect@7
Also connect@7 will force Cardano users to update FW to 2.0.11
connect@6 will be compatible with FW < 2.0.11

Issue #297, opened 2019-01-09
Flow: string literal window.close [1] is incompatible with object type [2].


SebastienGllmt:
If somebody uses `trezor-connect` in their node project with `flow-bin 0.86.0` you will get the error

```string literal window.close [1] is incompatible with object type [2].``` [here](https://github.com/trezor/connect/blob/5899a7f03741a00be933e7a1e4c0894bf88b081d/src/js/popup/PopupManager.js#L205)

PR #298, opened 2019-01-09
fixes for flow 86

SebastienGllmt:
This fixes Issue #297 and at the same time makes all the little changes to allow upgrading to flow 0.86.0


szymonlesisz:
cool, thanks!
i've additionally bump flow-bin to current version @0.91.0

PR #303, opened 2019-01-21
Added simplestaking.com to whitelist

jurajselep:


szymonlesisz:
as we speak thru email, we are not whitelisting 3rd party apps for web (popup) version

PR #308, opened 2019-01-29
Flash support

ashwinichourasia:
1) Code added to support Flashcoin in Trezor connect.
2) Once it merged with latest connect.js , Flashcoin developer team  can develope Wallet UI to support Flashcoin transaction using Trezor.
3) Flashcoin is already supported by latest Firmware of Trezor one and Trezor T.


prusnak:
Added in https://github.com/trezor/connect/commit/d3a9e8508ab74d05087b84c1b4ebb8a9fa265fc7

PR #320, opened 2019-02-07
Add management methods

mroz22:
opening discussion

- [ ] shouldt management methods have useEmptyPassphrase = true as default?
- [ ] camelCase/snake_case disorder. Connect does not have 100% clear convention which case should be used for calls input params. For example here https://github.com/trezor/connect/blob/65b4a064125e5e65fb5010b0fd0b43a75d06238d/src/js/core/methods/CardanoSignTransaction.js#L36 and here https://github.com/trezor/connect/blob/65b4a064125e5e65fb5010b0fd0b43a75d06238d/src/js/core/methods/ResetDevice.js#L31 

- [ ] naming - stick to protobuff definitions or make own names where it make sense? Just one case comes to my mind now - protobuff has `recoveryDevice`, nicer would be `recoverDevice`


src/js/core/methods/helpers/paramsValidator.js
```diff
@@ -6,9 +6,10 @@ import { fromHardened } from '../../../utils/pathUtils';
 import semvercmp from 'semver-compare';
 import type { CoinInfo } from '../../../types';
 
+// const types = 'string' | 'number' | 'array'
 type Param = {
     name: string,
-    type?: string,
+    type?: 'string' | 'number' | 'array' | 'array buffer' | 'boolean' | 'amount' | 'object',
```
szymonlesisz: make "arraybuffer" without space

src/js/core/Core.js
```diff
@@ -425,7 +425,17 @@ export const onCall = async (message: CoreMessage): Promise<void> => {
...
+            // todo
+            const unexpectedMode: ?(
+                typeof UI.BOOTLOADER |
+                typeof UI.INITIALIZE |
+                typeof UI.SEEDLESS |
+                typeof UI.FIRMWARE |
+                typeof UI.FIRMWARE_NOT_SUPPORTED
+            ) = device.hasUnexpectedMode(method.requiredFirmware, method.allowDeviceMode);
+
+            console.warn('unexpectedMode', unexpectedMode);
```
szymonlesisz: remove console

Makefile
```diff
@@ -27,8 +27,8 @@ build-connect:
 # Build test
 build-test:
 	make build-connect
-	sed -i '' -e 's/connect.trezor.io/sisyfos.trezor.io\/connect/g' ./build/js/iframe.*.js
-	rsync -avz --delete -e ssh ./build/* admin@dev.sldev.cz:~/sisyfos/www/connect
+	sed -i '' -e 's/connect.trezor.io/sisyfos.trezor.io\/martin-connect/g' ./build/js/iframe.*.js
+	rsync -avz --delete -e ssh ./build/* admin@dev.sldev.cz:~/sisyfos/www/martin-connect
```
szymonlesisz: remove custom build, just drop changes in this file

src/js/core/methods/ApplySettings.js
```diff
@@ -0,0 +1,41 @@
...
+            { name: 'language', type: 'string' },
+            { name: 'label', type: 'string' },
+            { name: 'use_passphrase', type: 'boolean' },
+            { name: 'homescreen', type: 'string' },
+            { name: 'passphrase_source', type: 'number' },
+            { name: 'auto_lock_delay_ms', type: 'number' },
+        ]);
+
+        this.params = {
+            label: payload.label,
```
szymonlesisz, line 40: why validating multiple fields when you are using only label? Or is it not finished?

src/js/core/methods/helpers/uploadFirmware.js
```diff
@@ -0,0 +1,36 @@
...
+export const uploadFirmware = async (
+    typedCall: (type: string, resType: string, msg: Object) => Promise<DefaultMessageResponse>,
+    payload: Buffer,
+    offset?: number,
+    length?: number,
+    model: number,
+): trezor.Success => {
+    let response: MessageResponse<trezor.Success | FirmwareRequest$> = {};
+
+    if (model === 1) {
```
szymonlesisz, line 16: i don't see this logic with strip 256 bytes for BL > 1.80? Shouldn't it be here?

src/js/core/methods/helpers/paramsValidator.js
mroz22: outdated, I have switched to plain buffer.

szymonlesisz:
good job!

PR #323, opened 2019-02-11
capricoin: added more tests

Adman:
Signed-off-by: Adrian Matejov <ado.matejov@gmail.com>

### DIFF
--- a/src/__tests__/core/signTransactionCapricoin.spec.js
+++ b/src/__tests__/core/signTransactionCapricoin.spec.js
@@ -52,9 +52,184 @@ const signCPC = (): SubtestSignTransaction => {
     };
 };
 
+const oneTwoFee = (): SubtestSignTransaction => {
+    // See tx 8302cb4b32815ac47e0d1a63081a7bbef843efeb7e29c414975f33dfe8b50e35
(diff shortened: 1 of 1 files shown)


Adman:
@szymonlesisz Hi, I've added some more test cases (except p2sh, because CPC does not support it), but I'm still unable to run these tests locally. Can you test it yourself, please?

```
LOG: '- Running /oneTwoFee ...'
WARN: Object{event: 'RESPONSE_EVENT', type: 'RESPONSE_EVENT', id: 1, success: false, payload: Object{error: 'Transport is missing'}}
ERROR: 'Cannot load debugLink state', 'Transport is missing'
ERROR: 'Unhandled promise rejection', TypeError: can't convert undefined to object
```

szymonlesisz:
Since capricoin doesn't support segwit shouldnt these two lines: https://github.com/trezor/trezor-common/blob/master/defs/bitcoin/capricoin.json#L17-L18 be set to null?

prusnak:
@szymonlesisz i will fix that, thanks

Adman:
@prusnak thanks for fixing it

@szymonlesisz is there anything more I can do?

PR #360, opened 2019-03-14
Add Komodo (KMD) coin to coins.json

smdmitry:
There is pull-request to bitcoinjs-lib-zcash to support Komodo transaction (https://github.com/karel-3d/bitcoinjs-lib/pull/13)


prusnak:
The correct way to add the coin is to change this file:

https://github.com/trezor/trezor-common/blob/master/defs/support.json

and add key `"bitcoin:KMD": true,` to `connect.supported`

Issue #369, opened 2019-03-27
Allow signTransaction if no backend is defined


prusnak:
We should allow the coins with no backends defined in their JSON because they still can be used via `signTransaction` if the caller provides the necessary info.

AlexITC:
From what I see on [SignTransaction.js](https://github.com/trezor/connect/blob/develop/src/js/core/methods/SignTransaction.js#L152), this is already supported, isn't it?

szymonlesisz:
already implemented here https://github.com/trezor/connect/commit/b0ec953bae95ff0b454cd6fc9050308783c96a92

PR #370, opened 2019-03-28
Allow to sign transactions without backend (#369)

JonSalazar:
Allow to sign self-contained transactions even if the coin has no backend

### DIFF
--- a/src/js/backend/BlockBook.js
+++ b/src/js/backend/BlockBook.js
@@ -51,9 +51,6 @@ export default class BlockBook {
     discovery: Discovery;
 
     constructor(options: Options) {
-        if (options.urls.length < 1) {
-            throw ERROR.BACKEND_NO_URL;
-        }
         this.options = options;
 
         const worker: FastXpubWorker = new FastXpubWorker();


szymonlesisz:
Looks easy, but unfortunately this is not the right place for this change.
If it not fail here it will fail in next step, inside `hd-wallet` (blockbook communication socket) and throw "Backend is offline" error.

To allow signing transaction for bitcoin-like coins without specified backend there must be several other changes:
1. add this coin to `trezor-common` and set support for connect
2. build coins.json from `trezor-common`
3. use `signTransaction` method with `refTxs` field set. There is already condition for not using backend at all if this field is set [here](https://github.com/trezor/connect/blob/develop/src/js/core/methods/SignTransaction.js#L145-L153)

AlexITC:
@szymonlesisz 

- Our coin is already in `trezor-common` with connect enabled (see https://github.com/trezor/connect/commit/d3a9e8508ab74d05087b84c1b4ebb8a9fa265fc7).
- From what I see, you already built the coins.json (https://github.com/trezor/connect/commit/d3a9e8508ab74d05087b84c1b4ebb8a9fa265fc7).
- The problem that I see on `signTransaction`, is that it will crash when there is no backend nor the reference transactions, we could add a conditional to enforce the reference transactions when there is no backend available, do you have a better idea?

We have already signed a transaction for XSN by providing all the reference transactions and the only change was what is included in this PR.

Thanks.

prusnak:
@AlexITC: Citing https://github.com/trezor/trezor-common/pull/302#issuecomment-477202697

> "from what I understand, we can call signTransaction passing all the arguments which avoids requiring any backend on the json"

So this is exactly what you need to do. If you want to use the backend functionality provided by the Connect, you need to provide the blockbook backend. That is all.

JonSalazar:
You're right, the change is unnecessary, thanks.

PR #372, opened 2019-04-02
Auto build dev branches in CI

vladimirvolek:


vladimirvolek:
Will be moved to monorepo.

PR #393, opened 2019-04-29
update MUE vals

sotblad:


szymonlesisz:
this is not the right place for this PR. This file is generated from:
https://github.com/trezor/trezor-common so your changes should be done there

PR #402, opened 2019-05-23
Flash Decimal fix

ashwinichourasia:
1) Fix is added to show Correct value of Flashcoin at time of sending 
We were facing a problem related to Satoshi decimals.
Generally Bitcoin and its Clones have
1 BTC = 100 000 000 Satoshis (8 Decimals)
1 LTC = 100 000 000 Litoshis. (8 Decimals)
But for FLASH,
its, 1 FLASH = 10 000 000 000 Satoshis (10 Decimals)
 So to display correct Value in send popup we need to divide it by 1e10 instead of 1e8

2) MinFee and MaxFee value for Flashcoin is increased.


prusnak:
Sorry, changing the decimals creates huge engineering costs on our side and this coin is not worth the effort. I am removing it completely from supported ones.

TheAkashKumar:
@prusnak Hi Pavol! I do understand your concerns. However, there are many Trezor users in the FLASH community who have been waiting to be able to use FLASH on their Trezor devices. I invite you to visit the FLASH community on Discord and you will realize that FLASH is worth it.

Moreover, we are in the process of decentralizing FLASH in the next few months and switching back to the standard 8 decimals would be part of it. At that point, it wouldn't require any changes from your side.

So kindly keep supporting FLASH on Trezor devices in the meanwhile.

prusnak:
When you switch back to 8 decimals, we'll put the FLASH back.

PR #404, opened 2019-05-27
Export input / output types for Cardano

SebastienGllmt:
I can't think of any reason why these should not be exported (I assume it was just an oversight).


szymonlesisz:
good catch! thanks!

PR #405, opened 2019-05-29
Update stellar types for stellar V11 (launching Monday June 3rd)

istrau2:
Adds additional types for Stellar v11 (going to production Monday June 3rd). Please see here:
https://github.com/stellar/js-stellar-base/blob/master/src/generated/stellar-xdr_generated.js#L2175

@fnando Do these changes look good to you? What about the legacy ops (i.e. manageOffer, createPassiveOffer), will those still work past Monday?

### DIFF
--- a/src/js/types/stellar.js
+++ b/src/js/types/stellar.js
@@ -48,8 +48,38 @@ export type ManageOfferOperation = {
     price: { n: number, d: number }, // Proto: "price_n" and "price_d"
 }
 
+export type ManageSellOfferOperation = {
+    +type: 'manageSellOffer', // Proto: "StellarManageOfferOp"
+    source: string, // Proto: "source_account"
+    offerId: number, // Proto: "offer_id"
+    amount: string, // Proto: parse to number
+    buying: Asset, // Proto: "buying_asset"
+    selling: Asset, // Proto: "selling_asset"
+    price: { n: number, d: number }, // Proto: "price_n" and "price_d"
+}
+
+export type ManageBuyOfferOperation = {
+    +type: 'manageBuyOffer', // Proto: "StellarManageOfferOp"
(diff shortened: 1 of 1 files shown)


szymonlesisz:
Hi @istrau2, 
Thanks for contribution, but those additional operations should be added to the firmware at the first place so Trezor device will be able to understand them. Adding types to trezor-connect is not enough.

istrau2:
@szymonlesisz Understood. This was more my way of introducing the issue (i.e. instead of just creating an issue, I thought I could start a PR).

I actually don't know how the upgrade to the network (planned for this Monday), will impact the current trezor sdk...

theaeolianmachine:
@szymonlesisz we have delayed the date until June 10th — what's the process of getting Trezor support when the network upgrades? The potential problems to come up seem to be:

* We added an additional operation (`ManageBuyOffer`), which if not supported users will not be able to utilize.
* We renamed two operations, which are are almost binary compatible but because of the change of offer IDs from uint64 to int64's, without changes this is likely to cause unexpected behavior.

We would highly advocate for making sure Trezor is compatible with Stellar V11's changes before the public network switches over on `2019-06-10 1600 UTC`.

Please feel free to [reach out to me on Keybase](https://keybase.io/aeolianmachine) with any questions you may have.

prusnak:
@theaeolianmachine Trezor firmware is released on the first Wednesday of a month and freezed two weeks prior that. Your changes will not fit in the June update, but they might fit in the July update if you send them soon enough. I suggest you send a pull request to our firmware monorepo: https://github.com/trezor/trezor-firmware before the next freeze. There are two types of firmware - legacy (T1) and core (T2), so please update both. Thank you!

prusnak:
Closing. As per the explanation above, the types need to be changed in the firmware repo first.

PR #418, opened 2019-07-15
allow fw updated when firmwareStatus: required

mroz22:
for discussion


src/js/device/Device.js
```diff
@@ -451,7 +451,7 @@ export default class Device extends EventEmitter {
     }
 
     isBootloader(): boolean {
-        return this.features.bootloader_mode && (typeof this.features.firmware_present === 'boolean' && this.features.firmware_present);
```
mroz22, line 454: Not 100% sure about this line, but I dont see readon for the deleted check.

mroz22: imho ok. was there on purpose but it doesnt last anymore

mroz22:
Should be ready for merge

PR #435, opened 2019-08-07
Polis info

eabz:
Missing Polis information.

### DIFF
--- a/src/data/coins.json
+++ b/src/data/coins.json
@@ -1751,6 +1751,50 @@
             "xpub_magic_segwit_p2sh": null
         },
         {
+            "address_type": 55,
+            "address_type_p2sh": 56,
+            "bech32_prefix": null,
+            "bip115": false,
+            "bitcore": [],
+            "blockbook": [
+                "https://blockbook.polispay.org"
+            ],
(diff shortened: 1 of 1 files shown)


szymonlesisz:
Hi @eabz,
Unfortunately this is not the right place for this PR.
`coins.json` is a file generated from [trezor-firmware/common](https://github.com/trezor/trezor-firmware/tree/master/common) respository.
You should add your definition to the /defs/bitcoin/ directory and also enable [support](https://github.com/trezor/trezor-firmware/blob/master/common/defs/support.json) for for connect

eabz:
Thanks @szymonlesisz 
But Polis def has been there for a long time: https://github.com/trezor/trezor-firmware/blob/master/common/defs/bitcoin/polis.json
Can you check why this isn't on the coins.json?

szymonlesisz:
it's because"POLIS" support is not added to "connect" https://github.com/trezor/trezor-firmware/blob/master/common/defs/support.json, its added only to the firmware support

eabz:
Thanks @szymonlesisz, I think I'll put an issue on trezor-firmware.

PR #449, opened 2019-09-05
Expose initial support for Hedera

mehcode:
Companion PR for https://github.com/trezor/trezor-firmware/pull/502

---

Even less sure how this repository is structured compared to **trezor/firmware**. Please let me know if there are any questions or concerns with the implementation and I'll work to correct them.


prusnak:
This coin does not seem relevant enough for addition.

PR #478, opened 2019-10-16
stellar plugin: support empty manageData value

MisterTicot:
While this is not part of the API documentation, manageData value may happen to be `undefined` for the purpose of deleting the account data entry from the ledger.

**Edit:** Added support for assets path

### DIFF
--- a/src/js/plugins/stellar/plugin.js
+++ b/src/js/plugins/stellar/plugin.js
@@ -30,16 +30,6 @@ const transformSigner = (signer) => {
 };
 
 /**
- * Transforms StellarSdk.Asset array to TrezorConnect.StellarTransaction.Asset
- * array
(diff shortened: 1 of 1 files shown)


src/js/plugins/stellar/plugin.js
```diff
@@ -134,6 +144,11 @@ const transformTransaction = (path, transaction) => {
             operation.signer = transformSigner(operation.signer);
         }
 
+        // transform asset path
+        if (operation.path) {
+          operation.path = transformPath(operation.path);
```
szymonlesisz: wouldn't be better to write `operation.path = operation.path.map(transformAsset)` instead of creating a new function?

MisterTicot: Absolutely.

PR #480, opened 2019-10-25
Tezos - Babylon update

jurajselep:


szymonlesisz:
1. there is no need to keep messages-default.json, please remove it
2. upadate tests fixtures: https://github.com/trezor/connect/blob/develop/src/__tests__/core/tezosSignTransaction.spec.js
(tests are failing)
2. since signTransaction params was changed please update docs as well (and examples with updated fixtures)

adonagy:
I've updated the tests and docs and also removed the unwanted files.

szymonlesisz:
great!

szymonlesisz:
released as:
https://www.npmjs.com/package/trezor-connect/v/8.0.8 
or
https://www.npmjs.com/package/trezor-connect/v/8.0.8-extended

PR #482, opened 2019-10-28
Fix/ripple lib error

szymonlesisz:
Temporary fix before `ripple-lib@1.3.5` release
[Referenced issue](https://github.com/ripple/ripple-lib/issues/1066) in ripple-lib


intelliot:
Should be fixed in ripple-lib@1.4.0

Issue #509, opened 2019-12-19
Can we make `TrezorConnect.getAddress` work with multisig addresses?


dhruvbansal:
The `TrezorConnect.getAddress` method does not respect `multisig` or `script_type` parameters when passed.  This prevents the Trezor device from displaying a multisig address, at least when being accessed through TrezorConnect.

* The [docs for TrezorConnect.getAddress](https://github.com/trezor/connect/blob/develop/docs/methods/getAddress.md) do not discuss the `multisig` nor `script_type` parameters.
* The [datatype definitions here](https://github.com/trezor/connect/blob/develop/src/data/messages/messages-v7.json#L572-L586) seem to suggest that `TrezorConnect.getAddress` can accept these parameters.
* But the [code for TrezorConnect.getAddress](https://github.com/trezor/connect/blob/develop/src/js/core/methods/GetAddress.js) just skips over them even if they are passed.

Trezor devices are certainly [capable of properly displaying multisig addresses](https://github.com/trezor/trezor-firmware/blob/08117331859edeaa210b8349b1c0e1efd6d0e23c/python/src/trezorlib/messages/GetAddress.py#L24-L25).  The Python API respects the `multisig` and `script_type` parameters which allows wallets such as Electrum to implement [multisig address display on Trezors](https://github.com/spesmilo/electrum/blob/428b63822b359d56d6ececabf406a43589545d24/electrum/plugins/trezor/trezor.py#L333-L346).

The TrezorConnect API supports `multisig` and `script_type` parameters in other contexts, for example when defining an input to a multisig address in `signTransaction`.

Adding this functionality would make Trezor devices much more secure for those accessing them through a web browser.

prusnak:
Hey Dhruv!

We just released connect 8.0.13 with your changes. Please test them out and report back whether everything works as expected.

Thank you for the contribution and @szymonlesisz for reviewing and adapting the changes!

PR #510, opened 2019-12-19
Make `getAddress` work with multisig BTC addresses.

dhruvbansal:
An attempt at solving https://github.com/trezor/connect/issues/509.

This PR attempts to correctly pass `multisig` and `scriptType` parameters in `TrezorConnect.getAddress` to lower-level code.  This is intended to enable displaying & confirming multisig BTC addresses from TrezorConnect. 

I have created this PR as a draft because it could benefit from a test case or few to test the new parameters are being passed correctly.  (I have been unable to run the test suite so far, still trying tho!)

### DIFF
--- a/src/js/device/DeviceCommands.js
+++ b/src/js/device/DeviceCommands.js
@@ -218,12 +218,13 @@ export default class DeviceCommands {
         return state;
     }
 
-    async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean): Promise<trezor.Address> {
-        const scriptType: trezor.InputScriptType = getScriptType(address_n);
+  async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean, multisig?: trezor.MultisigRedeemScriptType, scriptType?: trezor.InputScriptType): Promise<trezor.Address> {
+        if (!scriptType) { scriptType = getScriptType(address_n); }
         const response: Object = await this.typedCall('GetAddress', 'Address', {
             address_n,
             coin_name: coinInfo.name,
             show_display: !!showOnTrezor,
+            multisig,
             script_type: scriptType && scriptType !== 'SPENDMULTISIG' ? scriptType : 'SPENDADDRESS', // script_type 'SPENDMULTISIG' throws Failure_FirmwareError
         });
 
--- a/src/js/device/DeviceCommands.js
+++ b/src/js/device/DeviceCommands.js
@@ -218,7 +218,7 @@ export default class DeviceCommands {
         return state;
     }
 
-  async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean, multisig?: trezor.MultisigRedeemScriptType, scriptType?: trezor.InputScriptType): Promise<trezor.Address> {
+    async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean, multisig?: trezor.MultisigRedeemScriptType, scriptType?: trezor.InputScriptType): Promise<trezor.Address> {
         if (!scriptType) { scriptType = getScriptType(address_n); }
         const response: Object = await this.typedCall('GetAddress', 'Address', {
             address_n,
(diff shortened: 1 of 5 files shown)


src/js/device/DeviceCommands.js
```diff
@@ -218,12 +218,13 @@ export default class DeviceCommands {
...
-    async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean): Promise<trezor.Address> {
-        const scriptType: trezor.InputScriptType = getScriptType(address_n);
+    async getAddress(address_n: Array<number>, coinInfo: BitcoinNetworkInfo, showOnTrezor: boolean, multisig?: trezor.MultisigRedeemScriptType, scriptType?: trezor.InputScriptType): Promise<trezor.Address> {
+        if (!scriptType) { scriptType = getScriptType(address_n); }
         const response: Object = await this.typedCall('GetAddress', 'Address', {
             address_n,
             coin_name: coinInfo.name,
             show_display: !!showOnTrezor,
+            multisig,
             script_type: scriptType && scriptType !== 'SPENDMULTISIG' ? scriptType : 'SPENDADDRESS', // script_type 'SPENDMULTISIG' throws Failure_FirmwareError
```
szymonlesisz, line 228: it will not work because no matter what you set in `scriptType` this condition will exclude 'SPENDMULTISIG'
And the following comment is here for a reason :)

dhruvbansal: I was worried about this :)

I'm not 100% certain what `scriptType` even does downstream from here. Is it even necessary to achieve the goal of this PR (show multisig addresses on device from TrezorConnect)?  Or would it be sufficient to pass `multisig` downstream?

Thanks for your input!

szymonlesisz:
there is a reason for [this comment](https://github.com/trezor/connect/blob/beaacfbfcad5f5b642de9da8d0fcd178e48222a8/src/js/device/DeviceCommands.js#L228) "'SPENDMULTISIG' throws Failure_FirmwareError" :)

Maybe it was fixed in one of the latest releases and i didn't noticed that, need to investigate.
I know that its painful to run tests, we are working on it [here](https://github.com/trezor/connect/pull/503), but did you try to run [local version](https://github.com/trezor/connect/blob/develop/docs/index.md#running-local-version-developstable) of connect?

Anyway, i appreciate the effort :) I will investigate this issue soon, but there are more things to be done like exclude older not working firmwares.

dhruvbansal:
I'm trying to replicate via TrezorConnect the functionality that is already available via (say) the Python API and available in wallets such as Electrum.  This is what leads me to believe (hope?) that the firmware supports these additional parameters.

Perhaps I should just remove the guard on `scriptType` and let it pass through and see what happens.  I will attempt to get a local version of TrezorConnect working to test this.

I understand you are busy and I appreciate whatever time you can put into guiding me to land this PR!

src/js/device/DeviceCommands.js
szymonlesisz: Generally speaking `this.typedCall` will pass message to 'trezor-link' where it's going to be transformed to protobuf binary message and sent via transport (bridge > usb cable) to the device.

since both fields (multisig and script_type) are declared in ["GetAddress"](https://github.com/trezor/connect/blob/develop/src/data/messages/messages.json#L539-L580) this message should be accepted

szymonlesisz:
Try to run it locally, remove that problematic condition to see if it even works.
If it works, then we need to find out when it was fixed and add condition on top of "getAddress" function similar to [this one](https://github.com/trezor/connect/blob/develop/src/js/device/DeviceCommands.js#L277)
```
if (older_firmware) {
  exclude 'SPENDMULTISIG' like it was done before
} else {
  use 'SPENDMULTISIG' and  multisig field
}
```

szymonlesisz:
There is also a big chance that it wasn't firmware issue at all and it was connect bug (or rather lack of implementation) all along since no one was using it to export multisig addresses until now :)
that MultisigRedeemScriptType was not sent together with "script_type= SPENDMULTISIG"

szymonlesisz:
I can confirm that it works with little changes from me:
https://github.com/trezor/connect/commit/adf033fe2e611cfdd4a74f7223d8dc05f36e5439

I've used fixtures from [trezor-firmware](https://github.com/trezor/trezor-firmware/blob/4bae47ea6d760cb9bda7d3f388cdfe5e44e963df/tests/device_tests/test_msg_getaddress_show.py#L42)

and got same results, yay!

dhruvbansal:
Yay! Any timeline on when this would be released?  We'd love to expose this functionality to our users.

szymonlesisz:
next week i guess

prusnak:
Merged as part of https://github.com/trezor/connect/pull/516

PR #527, opened 2020-02-11
btc: add hex support to sign message.

nodech:
Add ability to optionally pass hex string to sign message. Also moved hex utilities to format utilities (I noticed they are generic, not ethereum related) This introduces a way to pass hex string instead of normal string for  signing binary messages. The API is same as for ethereumSignMessage,  where you have to pass optional hex parameter when you pass hex string in message. If hex is false, it will behave the same way as before.

---

*Accidentally removed branch from GH and it closed previous PR: #522*


szymonlesisz:
great! thanks for contribution

PR #548, opened 2020-03-29
Hotfix/v8/getAccountInfo default selected account

szymonlesisz:


src/js/core/methods/GetAccountInfo.js
```diff
@@ -62,6 +62,7 @@ export default class GetAccountInfo extends AbstractMethod {
                 { name: 'contractFilter', type: 'string' },
                 { name: 'gap', type: 'number' },
                 { name: 'marker', type: 'object' },
+                { name: 'defaultAccountType', type: 'boolean' },
```
prusnak: shouldn't this be `string` instead of `boolean`?

prusnak (changes requested):

src/js/core/methods/GetAccountInfo.js
szymonlesisz: good cath, fixed

PR #551, opened 2020-04-02
filter custom connectSrc

szymonlesisz:


src/js/data/ConnectSettings.js
```diff
@@ -74,6 +74,17 @@ export const getEnv = () => {
     return 'web';
 };
 
+// Cors validation copied from Trezor Bridge
+// see: https://github.com/trezor/trezord-go/blob/05991cea5900d18bcc6ece5ae5e319d138fc5551/server/api/api.go#L229
+// Its pointless to allow `trezor-connect` endpoints { connectSrc } for domains other than listed below
+// `trezord` will block communication anyway
+export const corsValidator = (url?: string) => {
+    if (typeof url !== 'string') return;
+    if (url.match(/^https:\/\/([a-zA-Z0-9-_])*.trezor.io\//)) return url;
```
prusnak: This should be `/^https:\/\/([A-Za-z0-9\-_]+\.)*trezor\.io\//`

in the original regex, the following did not work:
* `https://trezor.io/` (please add to test below)
* `https://sub.dom.trezor.io/` (please add to test below)

and this worked:
* `https://testxtrezorxio/`(please add to test below)

src/js/data/ConnectSettings.js
```diff
@@ -74,6 +74,17 @@ export const getEnv = () => {
...
+    if (url.match(/^https?:\/\/localhost:[58][0-9]{3}\//)) return url;
+    if (url.match(/^https:\/\/([a-zA-Z0-9-_])*.sldev.cz\//)) return url;
```
prusnak: This should be `/^https:\/\/([A-Za-z0-9\-_]+\.)*sldev\.cz\//`

same reason as above, please add to tests too

prusnak (changes requested):

PR #587, opened 2020-05-11
Fix Flow error in usb-permissions JS

SebastienGllmt:
This solves the two Flow errors in this file:

```
Cannot get `iframe.contentWindow` because property `contentWindow` is missing in  `HTMLElement` [1].Flow(InferError)
dom.js(1250, 38): [1] `HTMLElement`

Cannot get `iframe.contentWindow` because property `contentWindow` is missing in  null [1].Flow(InferError)
dom.js(1250, 52): [1] null
```


szymonlesisz:
nice, thanks!

PR #599, opened 2020-06-03
fix default FirmwareUpdate url

szymonlesisz:


prusnak:
This will break things because the paths differ:

https://wallet.trezor.io/data/firmware/2/trezor-2.3.0.bin (/data/...)
vs
https://data.trezor.io/firmware/2/trezor-2.3.0.bin (without /data/...)

In JSON we have the path with the "data/" at start, unfortunately. :-/

PR #633, opened 2020-07-27
Feat/timeout

keraf:
This adds a new configuration setting to set an interaction timeout. This timeout is triggered if the user doesn't interact with the popup or the device for a specified amount of time. The time can be set in seconds using the `interactionTimeout` property. By default, the interaction timeout is set to two minutes.


src/js/core/Core.js
```diff
@@ -58,6 +62,38 @@ const getPopupPromise = (requestWindow: boolean = true): Deferred<void> => {
...
 
+/**
+ * Start interaction timeout timer
+ */
+const interactionTimeout = () => _timeout.start(async () => {
+    /**
+     * Bridge version =< 2.0.28 has a bug that doesn't permit it to cancel
+     * user interactions in progress, so we have to do it manually.
+     */
+    const { version } = _core.getTransportInfo();
```
szymonlesisz: check only bridge transport (webusb not affected)
`const { type, version }`
`if (type === 'bridge' && versionCompare....`

src/js/utils/timeout.js
```diff
@@ -0,0 +1,58 @@
+/* @flow */
+
+import Log, { init } from './debug';
+
+const _log: Log = init('Timeout', true);
+
+export default class Timeout {
```
szymonlesisz: lets rename it to `InteractionTimeout` ? or `UITimeout`?

keraf: Yep, I'll rename it to InteractionTimeout for more clarity.

szymonlesisz (approved):

PR #623, opened 2020-07-07
Cardano Shelley Update 1/3

gabrielKerekes:
As mentioned earlier in #611, we are updating Cardano to support the new Shelley era.

Since it requires quite a lot of changes we are splitting it to (at least) 3 PRs on Trezor FW. We will do the same with Trezor Connect so they would stay in sync.

[Trezor FW Issue](https://github.com/trezor/trezor-firmware/issues/948)
[Trezor FW PR 1](https://github.com/trezor/trezor-firmware/pull/1096)

I've tried to find all the files that require changes, but I am not sure I found all the places that need to be updated.

I've updated types, methods and tests for both CardanoSignTransaction and CardanoGetAddress. Tests pass and I've also tried running it with an app running on localhost. I've also updated the READMEs, but I will check those once more when this gets approved. (In the READMEs I mention a mainnet protocol magic value of 0. We don't know what this value will be yet, so this will be updated when we find out.)

**I do have some questions regarding what should be updated:**
1. How do protobuf messages get updated? (I have been just copy and pasting `messages-cardano.proto` from Trezor FW repo)
2. Do both `cardanoSignTransaction.spec.js` and `tests/__fixtures__/cardanoSignTransaction.js` have to be updated? Or is the `.spec.js` osbolete now? (I'm going to be doing a lot more changes in the next PR and it would be simpler to edit just one - but it's not a big deal)
3. Are all the files edited manually or can they somehow be generated? (e.g. `src/ts/types/networks/cardano.d.ts` and `src/js/types/networks/cardano.js` are very similar and I was wondering if you edit them manually or not)

Please let me know if I have missed something major. Otherwise I'll be waiting for a normal review.

**As also mentioned in the Trezor FW PR it is possible that some things will change in the following days, but it should only be minor changes.**

PR #638, opened 2020-07-31
Cardano Shelley Update

gabrielKerekes:
As discussed in #623, this is a combination of 3 PRs - #623, https://github.com/vacuumlabs/connect/pull/2 and https://github.com/vacuumlabs/connect/pull/3. I sadly don't have time to somehow clean up the commits, but if you only check the end result, I think it shouldn't be that bad to review.


src/js/types/networks/cardano.js
```diff
@@ -19,42 +19,90 @@ export type CardanoPublicKey = {
 
 // GetAddress
 
-export type CardanoGetAddress = {
+export const CARDANO_ADDRESS_TYPE = Object.freeze({
```
szymonlesisz: this file should contains types only

pls move this const to `js/constants/cardano`

you can also put PROTOCOL_MAGICS and NETWORK_IDS there

src/__tests__/core/cardanoGetAddress.spec.js
```diff
@@ -1,51 +1,373 @@
 /* @flow */
 import type { CardanoGetAddress } from '../../js/types';
+import { CARDANO_ADDRESS_TYPE } from '../../js/types';
 
 // vectors from https://github.com/trezor/trezor-firmware/tree/master/python/trezorlib/tests/device_tests/test_msg_cardano_get_address.py
 
-export const cardanoGetAddress = () => {
+const PROTOCOL_MAGICS = {
+    mainnet: 0,
```
szymonlesisz: different values:

https://github.com/trezor/trezor-firmware/blob/c0eba979c6a1a5b6b54046c7d232743b6672291d/python/src/trezorlib/cardano.py#L22

using 0 will show "unknown network" on the device display

should be `764824073`

src/js/types/networks/cardano.js
gabrielKerekes: Moved [here](https://github.com/trezor/connect/pull/638/commits/bd67fef49aa1297383e7500cd07f567ce1c327eb).

src/__tests__/core/cardanoGetAddress.spec.js
gabrielKerekes: Oh yes, thanks. This has also been fixed by moving the constants [here](https://github.com/trezor/connect/pull/638/commits/bd67fef49aa1297383e7500cd07f567ce1c327eb).

src/__tests__/core/cardanoGetAddress.spec.js
```diff
@@ -1,51 +1,367 @@
 /* @flow */
+import {
+    CARDANO_ADDRESS_TYPE,
+    HARDENED,
```
szymonlesisz: does `HARDENED` need to be declared in constants since the only usage is only here in one test case? 
wouldn't be better to use `path: [0x80000000 + 49]` in line 29 ?

gabrielKerekes: Makes sense. Removed [here](https://github.com/trezor/connect/pull/638/commits/a272ece99493e82407d8fa6e283ab76990ea07ab).

prusnak:
The Travis check is failing, see https://travis-ci.org/github/trezor/connect/jobs/714401559

Is this because the tests are still on the old version of the Firmware? @szymonlesisz @mroz22 ?

szymonlesisz:
yes, test-env doesn't have 2.3.2 FW, we are working on that [here](https://github.com/trezor/connect/tree/feat/cardano-shelley-fork)

gabrielKerekes:
We have discovered some bugs while integrating with adalite, so the last two commits fix them. One was invalid path validation in address parameters - it was set to check that the type is string, which is not right for paths. And the second one was that withdrawals and certificates couldn't be empty. I'm sorry for the inconvenience.

szymonlesisz:
great,
i've cherrypicked those commits into my branch where i've made some additional changes, they are described here:
https://github.com/trezor/connect/pull/639

it should be merged in few moments and released tomorrow together with new firmware

prusnak:
Closed in favor of https://github.com/trezor/connect/pull/639

Thank you @gabrielKerekes for your contributions!

prusnak:
@gabrielKerekes Connect 8.1.10 released with your changes. It should provide Firmware Update 2.3.2 which includes your changes as well. You can try it out whether everything works as intended. Thank you again!

gabrielKerekes:
Nice! Everything seems to work fine with some basic use cases, we'll see what the users come up with.Thank you for your cooperation and for releasing everything including the late bug fixes!

PR #642, opened 2020-08-07
add new firmware 2.3.2 and 1.9.2

mroz22:
- [x] update releases.json

Tested local build with suite and did successfully update my device


szymonlesisz:
it is already there :) and already published as connect@8.1.10 on wednesday

PR #643, opened 2020-08-10
zcash: don't send version_group_id if version < 3

szymonlesisz:


prusnak:
Don't we also need to change this?

``` patch
diff --git a/src/js/core/methods/helpers/signtx.js b/src/js/core/methods/helpers/signtx.js
index 3758ee56..b9322fc7 100644
--- a/src/js/core/methods/helpers/signtx.js
+++ b/src/js/core/methods/helpers/signtx.js
@@ -55,7 +55,7 @@ const requestPrevTxInfo = (reqTx: RefTransaction,
             inputs_cnt: reqTx.inputs.length,
             outputs_cnt: outputCount,
             timestamp: reqTx.timestamp,
-            version_group_id: reqTx.version_group_id,
+            version_group_id: (reqTx.version >= 3) ? reqTx.version_group_id : null,
             expiry: reqTx.expiry,
             branch_id: reqTx.branch_id,
         };
```

or this code has already processed TX from `transformReferencedTransactions` where `version_group_id` is already correctly handled?

src/js/core/methods/tx/refTx.js
```diff
@@ -27,7 +27,7 @@ export const getReferencedTransactions = (inputs: Array<BuildTxInput>): Array<st
 export const transformReferencedTransactions = (txs: Array<BitcoinJsTransaction>): Array<RefTransaction> => {
     return txs.map(tx => {
         const extraData = tx.getExtraData();
-        const version_group_id = BitcoinJSCoins.isZcashType(tx.network) ? tx.versionGroupId : null;
+        const version_group_id = BitcoinJSCoins.isZcashType(tx.network) && typeof tx.versionGroupId === 'number' && tx.versionGroupId >= 3 ? tx.versionGroupId : null;
```
prusnak: This should be `tx.version >= 3` 

```suggestion
        const version_group_id = BitcoinJSCoins.isZcashType(tx.network) && typeof tx.versionGroupId === 'number' && tx.version >= 3 ? tx.versionGroupId : null;
```

prusnak (changes requested):

src/js/core/methods/tx/refTx.js
szymonlesisz: true that! thanks

PR #644, opened 2020-08-10
Add more CardanoGetAddress examples to readme

gabrielKerekes:
It wasn't obvious from the get address readme that also enterprise and
reward addresses are supported. So examples with these addresses have
been added.


szymonlesisz:
great, thanks!

PR #646, opened 2020-08-10
allow path 49 unhardened in pathUtils

logansease:
At Casa, we use unhardened paths for our multisig addresses, specifically 49/{0,1}/X/X/X
We use unhardened paths so that we can derive paths for multiple coin types and accounts on our server, without requiring the private key or user export of their device.
In Trezor-Connect v6, we have not had any problems with our multisig implementation, but in v8 we are having issues with our change addresses.
Specifically, the change output fails validation when we provide the device the node and script information required for the trezor device to authorize the the change output without user interaction.

Upon a deep dive into the trezor code, I see that this fails because the output validation does the following:

determine the script type for validation based on the Hardened path provided to the output
if the path found is not a hardened path of the types allowed, then the output script type is assumed to be 'PAYTOADDRESS'
the script for that output is then recomputed based on the information supplied and the type computed and compared against the provided script for validation.
The issue we are having is that we are using the script type 'PAYTOP2SHWITNESS', with unhardened paths. When the validation function determines the script type based on the hardened path, the unhardened path we provide does not fall into a case provided by the getScriptType function, and the default script type of 'PAYTOADDRESS' is used instead. When the validation function then computes the script based on PAYTOADDRESS, validation fails because the PAYTOADDRESS script computed does not match our provided PAYTOP2SHWITNESS script.

The solution I propose is quite simple. Rather than only checking for a hardened 49 path to get PAYTOP2SHWITNESS, if the raw (unhardened) path is 49 also return PAYTOP2SHWITNESS.

This will result in the getScriptType for our specific input returning PAYTOP2SHWITNESS which will then allow validation to validate with our expected script type and succeed.

PR #650, opened 2020-08-13
fix previous rules for Casa compatibility

prusnak:
See discussion in https://github.com/trezor/connect/pull/646#issuecomment-673617496 for more context

### DIFF
--- a/src/js/core/methods/helpers/signtxVerify.js
+++ b/src/js/core/methods/helpers/signtxVerify.js
@@ -68,13 +68,17 @@ const deriveBech32Output = (program: Buffer): Buffer => {
     return scriptSig;
 };
 
-const deriveOutputScript = async (getHDNode: GetHDNode, output: TransactionOutput, coinInfo: BitcoinNetworkInfo): Promise<Buffer> => {
+const deriveOutputScript = async (getHDNode: GetHDNode, output: TransactionOutput, coinInfo: BitcoinNetworkInfo): Promise<Buffer | void> => {
     if (output.op_return_data) {
         return BitcoinJsScript.nullData.output.encode(Buffer.from(output.op_return_data, 'hex'));
     }
     if (!output.address_n && !output.address) {
         throw ERRORS.TypedError('Runtime', 'deriveOutputScript: Neither address or address_n is set');
     }
+
+    // skip multisig output check, not implemented yet
+    // TODO: implement it
+    if (output.multisig) return;
 
     const scriptType = output.address_n
         ? getOutputScriptType(output.address_n)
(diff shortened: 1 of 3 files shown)


logansease:
Please see: https://github.com/trezor/connect/pull/652

src/js/core/methods/helpers/signtxVerify.js
```diff
@@ -68,18 +68,29 @@ const deriveBech32Output = (program: Buffer): Buffer => {
...
+const deriveOutputScript = async (getHDNode: GetHDNode, output: TransactionOutput, coinInfo: BitcoinNetworkInfo): Promise<Buffer | void> => {
     if (output.op_return_data) {
         return BitcoinJsScript.nullData.output.encode(Buffer.from(output.op_return_data, 'hex'));
     }
     if (!output.address_n && !output.address) {
         throw ERRORS.TypedError('Runtime', 'deriveOutputScript: Neither address or address_n is set');
     }
 
+    // compatibility for Casa - If passing in multisig params with p2shw, skip validation
+    if (output.multisig && output.script_type === 'PAYTOP2SHWITNESS') {
```
szymonlesisz: i would leave `if (output.multisig) return;`
like i said, script_type is irrelevant at this point

also check at line 90 could be removed

i'll do it

szymonlesisz:
@logansease pls verify if this branch works for you

src/js/utils/pathUtils.js
```diff
@@ -66,6 +66,11 @@ export const getScriptType = (path: ?Array<number>): InputScriptType => {
 export const getOutputScriptType = (path: ?Array<number>): OutputScriptType => {
     if (!Array.isArray(path) || path.length < 1) return 'PAYTOADDRESS';
 
+    // compatibility for Casa - allow an unhardened 49 path to use PAYTOP2SHWITNESS
+    if (path[0] === 49) {
+        return 'PAYTOP2SHWITNESS';
+    }
+
```
prusnak: Is this still needed? It seems to me that `getOutputScriptType` is not called anymore in Casa usecase (because of the early return in deriveOutputScript`).

szymonlesisz: yes, it is called on params validation (before signing, before sendin it to the device)
https://github.com/trezor/connect/blob/develop/src/js/core/methods/tx/outputs.js#L39

it was the first issue, and then there was a second issue: signed tx validation :)

szymonlesisz: so basically:
- connect checks if params are correct (solved by this condition)
- connect double-checks if signed tx is correct (solved by skip multisig check)

prusnak: thanks!

prusnak:
I added a changelog entry and rebased/squashed the branch on top of the current develop branch.

logansease:
Awesome @prusnak Thank you so much for your help. I have run this branch locally and confirmed signing works again as expected.

PR #652, opened 2020-08-14
revert script type change for casa to avoid validation with multisig + p2shw

logansease:
@prusnak 
Hey Pavol. Thank you for recommending me to run trezor locally so I could step through and test this. I hadn't been able to figure out how to do that previously.
I was able to test your proposed changes and they actually did not work for us. It was able to skip the output validation, however the signature did not apply on our end because the signature scripts did not match our script / address types since this required us to move to paytomultisig instead of using our current script types.

Since the issue you had mentioned was that p2shw ignored the multisig components during validation, the solution I am proposing instead is that we skip the script validation when this combination is passed in.
I have tested locally and confirmed this code now works. Would greatly appreciate you helping to move this through or letting me know if you have any other suggestions.

Thanks!

### DIFF
--- a/src/js/core/methods/helpers/signtxVerify.js
+++ b/src/js/core/methods/helpers/signtxVerify.js
@@ -77,7 +77,7 @@ const deriveOutputScript = async (getHDNode: GetHDNode, output: TransactionOutpu
     }
 
     // compatibility for Casa - If passing in multisig params with p2shw, skip validation
-    if (output.address_n && output.script_type === 'PAYTOP2SHWITNESS') {
+    if (output.multisig && output.script_type === 'PAYTOP2SHWITNESS') {
         return;
     }
 
--- a/src/js/core/methods/helpers/signtxVerify.js
+++ b/src/js/core/methods/helpers/signtxVerify.js
@@ -76,6 +76,11 @@ const deriveOutputScript = async (getHDNode: GetHDNode, output: TransactionOutpu
         throw ERRORS.TypedError('Runtime', 'deriveOutputScript: Neither address or address_n is set');
     }
 
+    // compatibility for Casa - If passing in multisig params with p2shw, skip validation
+    if (output.address_n && output.script_type === 'PAYTOP2SHWITNESS') {
+        return;
+    }
+
     const scriptType = output.address_n
         ? getOutputScriptType(output.address_n)
         : getAddressScriptType(output.address, coinInfo);
--- a/src/js/utils/pathUtils.js
+++ b/src/js/utils/pathUtils.js
@@ -50,11 +50,6 @@ export const isBech32Path = (path: ?Array<number>): boolean => {
 export const getScriptType = (path: ?Array<number>): InputScriptType => {
     if (!Array.isArray(path) || path.length < 1) return 'SPENDADDRESS';
 
-    // compatibility for Casa - allow an unhardened 49 path to use SPENDMULTISIG
-    if (path[0] === 49) {
-        return 'SPENDMULTISIG';
-    }
-
     const p1 = fromHardened(path[0]);
     switch (p1) {
         case 48:
@@ -70,11 +65,6 @@ export const getScriptType = (path: ?Array<number>): InputScriptType => {
 
 export const getOutputScriptType = (path: ?Array<number>): OutputScriptType => {
     if (!Array.isArray(path) || path.length < 1) return 'PAYTOADDRESS';
-
-    // compatibility for Casa - allow an unhardened 49 path to use PAYTOMULTISIG
-    if (path[0] === 49) {
-        return 'PAYTOMULTISIG';
-    }
 
     const p = fromHardened(path[0]);
     switch (p) {


prusnak:
Honestly, I don't like the change. It is too broad.

>  since this required us to move to paytomultisig instead of using our current script types.

Can you explain how this can be an issue for you? What are "your current script types"? If you want the output to behave as a change, it needs to be PAYTOMULTISIG. An explanation of what you are trying to achieve here would be very helpful - how does your standard transaction look like? What are the inputs and their scripts? What are the outputs leaving the wallet and their scripts? What are the change outputs and their scripts?

prusnak:
Please leave the comment in the original PR https://github.com/trezor/connect/pull/650 as I am closing this one.

logansease:
Thanks for re-opening to give this another look.

I'm trying to get lopp in on this, who has a better understanding of the different script types and how they work. My knowledge is limited. I do know that all of our address scripts are multisig pay to script hashes that are wrapped in pay to witness scripts (P2SH->P2WSH).

All of our inputs are from 3of5 multisig addresses which use the type P2SH->P2WSH and our change output is the same.
My specific goal is to pass the info required to trezor to avoid requiring the user to manually confirm the change output on their device. This works correctly on v6 since validation is skipped.

I also know that the code in the original PR that converted the output to use paytomultisig generated signatures properly, but that our backend was unable to validate and accept them because it also validates the scripts of the type P2SH->P2WSH. So, I think that when the code above converted the output script type to paytomultisig something was lost. Perhaps they are no longer being wrapped in witness scripts / segwit? 

From your code, I also tried various permutations of PAYTOP2SHWITNESS vs PAYTOMULTISIG on both our side and in the getScriptType method from this PR and none of the permutations were able to generate a signature that was accepted by both trezor and our backend.

Note that I don't have problems with the SPENDP2SHWITNESS on the input scripts in the current code when there is not a change output. It is only when we have a change output and try to avoid requiring the user to validate the output on their device manually by specifying PAYTOP2SHWITNESS and providing the multisig info. 

Here is an example transaction in case it is helpful:
```
010000000001015fe4828f7ce8c4158e07fe30d9145477dd3462dda6a92f53d7d284ec9605f6e7010000002322002005952df701bd03165b5055be6e03acf949cc2648756d30b0edc6500d22747e6cffffffff022a4200000000000017a914823ca2ab3ff7f73f3b713872a83777a499090c6387a5741f000000000017a91473d45b78f0c06e9abb4ff392b8bb163df0f245778707000000000000ad532102198b91e17a7588044d95b0259e502dc4827fec2e6e1a103746acdedc08593de221024561c2412c7e196bcb47dc4880b98c8baf41711a17ac2a94f130213ea6c2da5b2102b9a9e3eb6d1a2ef6101342dd2005f975410ea0d018eab7673a6a836d6e3d919d2103bdc2a68310828fe1899805c57e05c3b7b67f280037c78aa528b2f819265364392103fe4d79bdb35dc21659f688362102cb5b23d6f24ac2b7c107811761c9ded7b01255ae00000000
```

```
Decoded: 
{
    "addresses": [
        "2MszAT2c7K6fz887c5QPvXGevbbvb4yKnez",
        "2N57rUFAtKyZCkFiydBCDBg4675jT1iFJyQ",
        "2N3og6Bw5tYzK2e4YPXo9EKbuo37iffTbEE"
    ],
    "block_height": -1,
    "block_index": -1,
    "confirmations": 0,
    "double_spend": false,
    "fees": 578,
    "hash": "c3ab176b4bf7bdfbbb23b30438d407db7e8fd56314a51c8bdeebf2d5581e87e6",
    "inputs": [
        {
            "addresses": [
                "2MszAT2c7K6fz887c5QPvXGevbbvb4yKnez"
            ],
            "age": 1780354,
            "output_index": 1,
            "output_value": 2078993,
            "prev_hash": "e7f60596ec84d2d7532fa9a6dd6234dd775414d930fe078e15c4e87c8f82e45f",
            "script": "22002005952df701bd03165b5055be6e03acf949cc2648756d30b0edc6500d22747e6c",
            "script_type": "pay-to-script-hash",
            "sequence": 4294967295
        }
    ],
    "outputs": [
        {
            "addresses": [
                "2N57rUFAtKyZCkFiydBCDBg4675jT1iFJyQ"
            ],
            "script": "a914823ca2ab3ff7f73f3b713872a83777a499090c6387",
            "script_type": "pay-to-script-hash",
            "value": 16938
        },
        {
            "addresses": [
                "2N3og6Bw5tYzK2e4YPXo9EKbuo37iffTbEE"
            ],
            "script": "a91473d45b78f0c06e9abb4ff392b8bb163df0f2457787",
            "script_type": "pay-to-script-hash",
            "value": 2061477
        }
    ],
    "preference": "low",
    "received": "2020-08-14T21:49:02.113370627Z",
    "relayed_by": "54.175.124.1",
    "size": 150,
    "total": 2078415,
    "ver": 1,
    "vin_sz": 1,
    "vout_sz": 2
}
```

szymonlesisz:
tbh i think the best solution is to skip validation for every output containing multisig regardless of it script_type.
at this point it doesn't matter if its `PAYTOMULTISIG` or `PAYTOP2SHWITNESS` or else. It will always fail because generating script output for multisig is not implemented (yet)

prusnak:
OK. I'll merge this into the PR #650 and then rebase on top of the current develop + apply Szymon's suggestion. Please have a look at the update #650 afterwards ...

PR #660, opened 2020-08-20
Fix/cardano double passphrase

### LINKED ISSUE #593: Provide option to skip the passphrase check for Cardano
In https://github.com/trezor/trezor-firmware/issues/998 we have found out that the way we currently check the correctness of passphrase via GetAddress has an ugly consequence for Cardano.

Since Cardano uses different seed derivation mechanism in Trezor and we do not store the passphrase in firmware in plaintext we need to request it again. This leads to a duplicate passphrase prompt on Trezor.

We could solve this in Connect by providing option to skip this passphrase check. @szymonlesisz what do you think?

This has low priority.

cc @matejcik

keraf:
Added `network` property to `AbstractMethod` base class. Added network parameter to `getDeviceState` method in `DeviceCommands` class so it can send the right call to the device and prepare the correct seed derivation mechanism (thus preventing the second passphrase request).

### DIFF
--- a/src/ts/types/constants.d.ts
+++ b/src/ts/types/constants.d.ts
@@ -170,7 +170,7 @@ export namespace CARDANO {
     }
 }
 
-export namespace NETWORK {
+declare namespace NETWORK {
     enum TYPES {
         'bitcoin' = 'Bitcoin',
         'etherum' = 'Etherum',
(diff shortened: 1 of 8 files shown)


src/ts/types/constants.d.ts
```diff
@@ -169,3 +169,18 @@ export namespace CARDANO {
         StakeDelegation = 2,
     }
 }
+
+export namespace NETWORK {
```
szymonlesisz: you don have to export those, it's a internal const

PR #662, opened 2020-08-22
coins: Add Decred and Decred Testnet.

JoeGruffins:
This adds Decred mainnet and testnet coin parameters.


src/data/coins.json
```diff
@@ -1780,6 +1780,92 @@
...
+            "minfee_kb": 10000,
+            "name": "Decred",
+            "segwit": false,
+            "shortcut": "DCR",
+            "signed_message_header": "Decred Signed Message:\n",
+            "slip44": 42,
+            "support": {
+                "connect": true,
+                "trezor1": "1.8.0",
+                "trezor2": "2.0.11",
```
JoeGruffins: These are probably not accurate. I will keep snooping around to find the correct values, unless it's obvious?

JoeGruffins:
Removed simnet as that will not be going into the firmware.

src/data/coins.json
JoeGruffins: Using values found here:
https://github.com/trezor/trezor-firmware/blob/b00c4209eda71e6482fa3cea5c0d14d648f474f5/common/defs/support.json#L106

JoeGruffins:
Not sure what happened, but looks like decred stuff is merged, closing.

PR #666, opened 2020-08-26
Feat/getpubkey showtrezor

### LINKED ISSUE #401: feature: add "show on device" for getPublicKey
I'd like to be able to show users their xpubs with verification on-device.

keraf:
- Adds a parameter to confirm the public key on device (`showOnTrezor`) for the GetPublicKey method.
- Wired up showOnTrezor for EthereumGetPublicKey (was previously always false regardless of the passed parameter).

### DIFF
--- a/src/js/device/DeviceCommands.js
+++ b/src/js/device/DeviceCommands.js
@@ -106,11 +106,13 @@ export default class DeviceCommands {
         address_n: Array<number>,
         coin_name: string,
         script_type?: ?string,
+        show_display?: ?boolean,
     ): Promise<trezor.PublicKey> {
         const response: MessageResponse<trezor.PublicKey> = await this.typedCall('GetPublicKey', 'PublicKey', {
             address_n,
             coin_name,
             script_type,
+            show_display,
         });
         return response.message;
     }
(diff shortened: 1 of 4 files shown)


src/js/device/DeviceCommands.js
```diff
@@ -144,11 +147,11 @@ export default class DeviceCommands {
 
         let publicKey: trezor.PublicKey;
         if (!validation) {
-            publicKey = await this.getPublicKey(path, coinInfo.name, scriptType);
+            publicKey = await this.getPublicKey(path, coinInfo.name, scriptType, showOnTrezor);
```
szymonlesisz, line 150: skip validation if showOnTrezor is set, so:
if (showOnTrezor || !validation)

src/js/device/DeviceCommands.js
```diff
@@ -144,11 +147,11 @@ export default class DeviceCommands {
...
         } else {
             const suffix: number = 0;
             const childPath: Array<number> = path.concat([suffix]);
-            const resKey: trezor.PublicKey = await this.getPublicKey(path, coinInfo.name, scriptType);
+            const resKey: trezor.PublicKey = await this.getPublicKey(path, coinInfo.name, scriptType, showOnTrezor);
```
szymonlesisz: remove showOnTrezor param, it will never happen if showOnTrezor = true

szymonlesisz:
+ update changelog and docs

PR #671, opened 2020-08-28
multi: Add decred transaction validation. 

JoeGruffins:
This depends on #662 and https://github.com/trezor/trezor-utxo-lib/pull/9

Adds some needed support for signing and verifying transactions for Decred.


JoeGruffins:
Not sure what happened, but looks like decred stuff is merged, closing.

PR #672, opened 2020-08-28
fix T1 + bootloader < 1.4.0

szymonlesisz:
T1 + bootloader < 1.4.0 doesn't know the "GetFeatures" message yet and it will send no response to it
transport response is pending endlessly, calling any other message will end up with "device call in progress"

set the timeout for this call so whenever it happens "unacquired device" will be created instead
next time device should be called together with "Initialize" message (by calling "acquireDevice" from the UI)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 8.1.13 (not released)
+
+### Fixed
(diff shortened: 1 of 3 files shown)


keraf (approved):

keraf:
Could you update the changelog as well please?

PR #685, opened 2020-09-24
fix: cancel method called during Pin/Passphrase/Word requests

szymonlesisz:
fix/patch for TrezorConnect.cancel method, see description in in file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # 8.1.15 (unreleased)
 
+### Fixed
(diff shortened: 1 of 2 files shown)


bosomt:
**QA OK**

Works fine now

Nixos + Chromium 85
Suite commit 6fa440f1fd278ceb66341afc47872c2b2cf60f39 + custom coneect via URL
FW 2.3.3 df5421e

keraf (approved):

PR #692, opened 2020-10-12
chore: add trezor onion address to allowed list

szymonlesisz:


src/js/data/ConnectSettings.js
```diff
@@ -84,6 +84,7 @@ export const corsValidator = (url?: string) => {
     if (url.match(/^https:\/\/([A-Za-z0-9\-_]+\.)*trezor\.io\//)) return url;
     if (url.match(/^https?:\/\/localhost:[58][0-9]{3}\//)) return url;
     if (url.match(/^https:\/\/([A-Za-z0-9\-_]+\.)*sldev\.cz\//)) return url;
+    if (url.match(/^https?:\/\/trezoriovpjcahpzkrewelclulmszwbqpzmzgub37gbcjlvluxtruqad\.onion\//)) return url;
```
keraf: We should allow all subdomains, similarly to trezor.io.

szymonlesisz: taken from the trezord repo https://github.com/trezor/trezord-go/commit/9a7e757ce1b03054b0d2da9aa660b324a12a764c

keraf: According to trezor/trezor-suite#2445 we'll use it from suite.trezorio[...].onion (not sure if we want to include other subdomains). I'm afraid it might break on trezord.

szymonlesisz: yes, you're right! it needs to be fixed in trezord as well

PR #699, opened 2020-10-25
Introduce Gitlab

tsusanka:
This is not perfect as I have finally not changed the run.sh script, but I would like to merge this. What can be done later:

- Use Trezor User Env also in Travis.
- Introduce new run.sh script for local run. I am not sure this is needed as one can run trezor-user-env on their own. We can discuss it later.


tsusanka:
Not sure one of the Travis fails, if the other succeeds :shrug: . Should be fine on restart.

PR #708, opened 2020-11-13
Fix/test with fw 2.3.4

szymonlesisz:
#### this PR should be squashed into 1 commit
fix https://github.com/trezor/connect/issues/698

PR #709, opened 2020-11-13
Rewrite run.sh script to allow running tests locally

tsusanka:
Change run-travis.sh to universal run.sh that can be run locally. Uses new trezor-user-env. Things to do:

- [ ] Check on Mac. I don't have one at the moment. Should work also with GUI if `-e DISPLAY=:0` is replaced with `-e DISPLAY=docker.for.mac.host.internal:0`.
- [ ] Check if Travis works. For some reason [it does not run](https://travis-ci.org/github/trezor/connect/builds/743466649) now but develop does not run either so it is probably not related to this PR.


tests/run.sh
```diff
@@ -0,0 +1,119 @@
...
+  echo ""
+  echo "Options:"
+  echo "  -g       Run tests with emulator graphical output"
+  echo "  -f       Use specific firmware version, for example: 2.1.4., 2.3.0"
+  echo "  -i       Included methods only, for example: applySettings,signTransaction"
+  echo "  -e       All methods except excluded, for example: applySettings,signTransaction"
+  echo "  -c       Collect coverage"
+}
+
+FIRMWARE='2.3.2'
```
szymonlesisz: 2.3.4

tests/run.sh
```diff
@@ -0,0 +1,119 @@
...
+    # TODO: I don't think these options are needed:
+    # -v /tmp/.X11-unix:/tmp/.X11-unix:rw \
+    # -e QT_X11_NO_MITSHM=1 \
+
+    echo "Running docker container with a GUI support with ID:"
+    echo "$id"
+  fi
+
+  echo "Waiting for the trezor-user-env to load up"
+  sleep 10
```
szymonlesisz: not sure about that delay. it may cause some unexpected race conditions. isn't there other way to catch trezor-user-env load event?

szymonlesisz (commented):
rebase this branch on top of https://github.com/trezor/connect/pull/708
failing test are fixed there.

i've tried it on mac, but its failing on websocket connecton. im blaming docker process command, because if i run it as standalone the output is:
```
> docker run -it -e SDL_VIDEODRIVER="dummy" -p "21326:21326" -p "9001:9001" -p "21325:21325" registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:latest /trezor-user-env/run.sh
> Python 3.8.5
> trezorctl, version 0.12.2
> Starting trezor-user-env server
>
```

tests/run.sh
tsusanka: f7dc340

tests/run.sh
tsusanka: Yeah probably yes, I was just lazy. I'll have a look.

tsusanka: https://github.com/trezor/connect/pull/709/commits/0e2c2830795d3bd75ee6c4fcb9c38b4be6bd4999

tsusanka:
Please test https://github.com/trezor/connect/pull/709/commits/0e2c2830795d3bd75ee6c4fcb9c38b4be6bd4999 on Mac and if it works I can squash and we'll merge.

szymonlesisz (approved):
it's failing on zcash tests as expected, they are fixed in other branch

PR #730, opened 2020-12-18
Fix inconsistent acc. number between cardanoGetPublicKey and cardanoG…

refi93:
Motivation: When exporting the Cardano public key, connect shows the account number as `#1` (as well as for other currencies), but for the address verification call it shows the same account as `#0` (i.e. does not increase the account number). This is inconsistent behavior which makes sense to align with the rest of coins.

Changes: Increase displayed account number by 1 for CardanoGetAddress

Testing: tested locally end-to-end, works as expected


szymonlesisz:
great! looks legit, thanks

PR #742, opened 2021-01-13
fix: use preferred backend connection

szymonlesisz:
Try to use preferred backend connection
Switching between backends during application (suite) lifetime could lead to side effects where mempools are not synchronized properly and backend-A is showing pending transaction while backend-B doesnt have it

Try to stick with one url once connection is succesfully made

### DIFF
--- a/src/js/backend/BlockchainLink.js
+++ b/src/js/backend/BlockchainLink.js
@@ -90,6 +90,9 @@ export default class Blockchain {
                 return;
             }
 
+            // eslint-disable-next-line no-use-before-define
+            setPreferredBacked(this.coinInfo, info.url);
+
             this.postMessage(BlockchainMessage(BLOCKCHAIN.CONNECT, {
                 coin: this.coinInfo,
                 ...info,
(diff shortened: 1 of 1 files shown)


src/js/backend/BlockchainLink.js
```diff
@@ -289,7 +305,7 @@ export const initBlockchain = async (coinInfo: CoinInfo, postMessage: $ElementTy
     let backend = find(coinInfo.name);
     if (!backend) {
         backend = new Blockchain({
-            coinInfo: customBackends[coinInfo.shortcut] || coinInfo,
+            coinInfo: customBackends[coinInfo.shortcut] || preferredBackends[coinInfo.shortcut] || coinInfo,
```
keraf: Shouldn't this be `coinInfo: preferredBackends[coinInfo.shortcut] || customBackends[coinInfo.shortcut] || coinInfo,` ?

szymonlesisz: ha, of course it should!
fixed, thx

keraf (approved):

PR #745, opened 2021-01-25
Add Cardano multiasset support

refi93:
Motivation: Trezor connect updates needed to support multiasset/validity interval start in Cardano

Depends on https://github.com/trezor/trezor-firmware/pull/1432

Testing:
* integration tests with local trezor-firmware build based on https://github.com/trezor/trezor-firmware/pull/1432 : `./tests/run-travis.sh -b ~/workplace/cardano/trezor-firmware/core/build/unix/trezor-emu-core-v2.9.9 -f 2.9.9 -i cardanoSignTransaction` pass
* `yarn flow` passes
* `yarn eslint` passes

Note: The PR is currently based on 966f51d commit from connect's develop branch, as on the current develop branch tip (https://github.com/trezor/connect/commit/5f40141ee67ce51501f48514a2356a6b29b22fb9) I was unable to get the integration tests working with custom trezor firmware and the docs on that seem to be outdated. Here's the issue I raised where I'm seeking guidance: trezor#723

PR #753, opened 2021-02-01
Validate firmware version for newly added fields in CardanoSignTransaction

### LINKED ISSUE #751: Preventing forward compatibility issues
In Cardano we recently added multiasset support/minValidity field (https://github.com/trezor/connect/pull/745) which extends the existing protobuf messages with new fields. This works nice with old trezor-firmware (2.3.4) until a user attempts to sign a transaction containing those new fields. The expected behavior would be that Trezor rejects such transaction as it contains unknown fields that it cannot serialize into the tx, but instead (due to the way protobuf is designed) it proceeds to gracefully sign a transaction without those fields which is invalid at best and lacking some crucial information at worst (e.g. the min validity start field, effectively disabling the time lock on the signed tx).

Is there any better way to prevent such forward compatibility-related issues than explicitly checking the trezor firmware version client-side? Am I perhaps missing some pattern/helper function inside trezor-connect where we could specify since which version of Trezor firmware does given field work and throw if a message with that field is passed?

I guess it's loosely related to https://github.com/trezor/trezor-firmware/issues/1438

refi93:
Motivation: In Cardano we recently added multiasset support/minValidity field (#745) which extends the existing protobuf messages with new fields. This works nice with old trezor-firmware (2.3.4) until a user attempts to sign a transaction containing those new fields. The expected behavior would be that Trezor rejects such transaction as it contains unknown fields that it cannot serialize into the tx, but instead (due to the way protobuf is designed) it proceeds to gracefully sign a transaction without those fields which is invalid at best and lacking some crucial information at worst (e.g. the min validity start field, effectively disabling the time lock on the signed tx).

Changes:
* add checks for firmware version to prevent previous Trezor firmware versions from leading the user to accidentally sign a transaction with those fields missing, giving a comprehensive error message
* fix tokenBundle validation (validating the non-existing snake_case property instead of camelCase)

Another valid approach would be globally forcing trezor firmware version 2.3.5 for cardanoSignTransaction but given the relatively short window between it's expected launch (Feb 10th) and the upcoming hard fork in Cardano (~Feb 26th) it's probably more sensible to have sooner a version of connect working both with 2.3.4 and 2.3.5 at the same time and enforce 2.3.5 (or something higher) globally later on with another batch of less critical changes.

Testing:
* `./tests/run.sh -i cardanoSignTransaction -g` passes
* `./tests/run.sh -i cardanoSignTransaction -g -f 2.3.4` fails with the fixtures concerning changes added after fw version 2.3.4 was launched, as expected

closes https://github.com/trezor/connect/issues/751

### DIFF
--- a/src/js/core/methods/CardanoSignTransaction.js
+++ b/src/js/core/methods/CardanoSignTransaction.js
@@ -7,6 +7,10 @@ import { validatePath } from '../../utils/pathUtils';
 import { addressParametersToProto, validateAddressParameters } from './helpers/cardanoAddressParameters';
 import { transformCertificate } from './helpers/cardanoCertificate';
 import { validateTokenBundle, tokenBundleToProto } from './helpers/cardanoTokens';
+import { ERRORS } from '../../constants';
+import Device from '../../device/Device';
+import { CERTIFICATE_TYPE } from '../../constants/cardano';
+
 
 import type {
     MessageType,
@@ -16,6 +20,13 @@ import type {
     CardanoTxWithdrawalType,
 } from '../../types/trezor/protobuf';
 import type { CoreMessage } from '../../types';
+
+// todo: remove when listed firmwares become mandatory for cardanoSignTransaction
+const CardanoSignTransactionFeatures = Object.freeze({
+    SignStakePoolRegistrationAsOwner: ['', '2.3.5'],
+    ValidityIntervalStart: ['0', '2.3.5'],
+    MultiassetOutputs: ['0', '2.3.5'],
+})
 
 export default class CardanoSignTransaction extends AbstractMethod {
     params: $ElementType<MessageType, 'CardanoSignTx'>;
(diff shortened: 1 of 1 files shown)


src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -114,7 +125,35 @@ export default class CardanoSignTransaction extends AbstractMethod {
...
+    _ensureFirmwareSupportsParams() {
+        const params = this.params;
+
+        params.certificates.map((certificate) => {
+            if (certificate.type === CERTIFICATE_TYPE.StakePoolRegistration) {
+                this._ensureFeatureIsSupported('SignStakePoolRegistrationAsOwner');
+            }
+        });
+
+        if (params.validity_interval_start != null) {
```
szymonlesisz, line 141:
```suggestion
        if (typeof params.validity_interval_start === 'number') {
```
!= null does not exclude `undefined` case

src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -17,6 +21,13 @@ import type {
 } from '../../types/trezor/protobuf';
 import type { CoreMessage } from '../../types';
 
+// todo: remove when listed firmwares become mandatory for cardanoSignTransaction
+const CardanoSignTransactionFeatures = Object.freeze({
+    SignStakePoolRegistrationAsOwner: ['', '2.3.5'],
```
szymonlesisz:
```suggestion
    SignStakePoolRegistrationAsOwner: ['0', '2.3.6'],
```
0 to be consistent with others and `2.3.6` - because 2.3.5 was never released

szymonlesisz (changes requested):
looks good, just few nitpicks

src/js/core/methods/CardanoSignTransaction.js
refi93: not sure if I got you, the purpose of the condition is to not trigger when `params.validity_interval_start` is either null or undefined. The following snippet

```
const x = {}
console.log(x.foo != null)
```
prints `false` which is desired. `undefined ` would be left out if the operator was `!==`

szymonlesisz: ah, now i see it `validityIntervalStart` is a string (number represented as string, right?)
maybe it should be validated properly in line 39 (as 'amount' instead of 'string')?

otherwise im able to set param validityIntervalStart: 'foo-bar'

refi93: Initially I set it to string as it didn't feel right to mix it with true amounts but validation-wise they are the same, so I agree. Maybe an alias for `uint64` for generic unsigned big numbers in paramsValidator would be nice, but it's not a big deal after all

szymonlesisz (approved):

szymonlesisz:
squashed and merged (i needed to run test locally since travis is not working anymore)

src/js/core/methods/CardanoSignTransaction.js
refi93: @szymonlesisz - I see changelog still lists the version 2.3.5 as the one to be released on February 10th: https://github.com/trezor/trezor-firmware/blob/44c7d2374168fab867e167095c0e32aa5fdcea39/core/CHANGELOG.md - is that outdated?

PR #757, opened 2021-02-02
fix: faulty check for token bundle presence in Cardano protobuf message

refi93:
Motivation: the check for presence of token_bundle in tx outputs turned out to be incorrect as it gets translated to an empty array which is truthy.Therefore Trezor fails with `Feature MultiassetOutputs not supported by device firmware` error for any transaction, not just those actually containing tokens. This is a fix for this by checking explicitly for the length of the array

Testing:
* `./tests/run.sh -i cardanoSignTransaction -g -f 2.3.4` fails with the 2.3.6 specific features only and the rest passes
* `./tests/run.sh -i cardanoSignTransaction -g` passes

@szymonlesisz - this is urgent as it started affecting adalite users given that the connect instance deployed on the web already has the newest version with this bug, sorry for the hassle

### DIFF
--- a/src/js/core/methods/CardanoSignTransaction.js
+++ b/src/js/core/methods/CardanoSignTransaction.js
@@ -143,7 +143,7 @@ export default class CardanoSignTransaction extends AbstractMethod {
         }
 
         params.outputs.map((output) => {
-            if (output.token_bundle) {
+            if (output.token_bundle.length > 0) {
                 this._ensureFeatureIsSupported('MultiassetOutputs');
             }
         });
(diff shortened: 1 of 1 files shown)


src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -143,7 +143,7 @@ export default class CardanoSignTransaction extends AbstractMethod {
         }
 
         params.outputs.map((output) => {
-            if (output.token_bundle) {
+            if (output.token_bundle.length > 0) {
```
szymonlesisz:
```suggestion
            if (output.token_bundle && output.token_bundle.length > 0) {
```
token_bundle is optional right?

refi93: technically, it's a "repeated field" at protobuf level and it gets translated to an empty array above even if it is missing, but I agree for the sake of "fool proofness" better explicitly check for its presence

szymonlesisz:
deployed

PR #764, opened 2021-02-15
chore: use FW 2.3.5 in tests

szymonlesisz:


szymonlesisz:
on second thought it's better to have them set to firmware master

PR #766, opened 2021-02-19
feat(cardano): add support for chunked tx serialization

refi93:
Motivation: extend cardano tx serialization to be chunked, so the whole serialized tx doesn't need to be held in-memory (which was causing serious throughput issues)

This changes are backwards compatible with Trezor fw 2.3.5 and older as only a "expect_more_chunks" flag has been added at protobuf level. However, older versions of connect won't work with Trezor firmware updated with this extension as they would receive only the first chunk of the transaction

Related PR with further explanation: https://github.com/trezor/trezor-firmware/pull/1493

Testing:
* `yarn test:integration -i cardanoSignTransaction -g -f 2.3.6` passes, also I tested it with the firmware built from the PR referenced above by monkey-patching the trezor-user-env docker container


gabrielKerekes:
PR has been updated to correspond with applied review suggestions from the related firmware PR.

gabrielKerekes:
Once again updated to reflect changes in firmware PR. The ordering of witnesses in transactions has been changed so some tests needed to be updated.

gabrielKerekes:
I have squashed and rebased on top of current develop. `cardanoSignTransaction` tests no longer pass with older firmware since the witnesses order has been changed in some transactions, but the tests do pass with FW from the related PR.

szymonlesisz:
merged via https://github.com/trezor/connect/pull/790/commits

i've added possibility to use [legacy results in tests](https://github.com/trezor/connect/pull/790/commits/9bf4db8faab7be40356cad8ca44b4cdc3e5a142a) for lower FW version and restored your recent changes (tested locally on 2.3.6 and 2-master)

PR #771, opened 2021-03-01
fix: set log messages limit

szymonlesisz:
- set limit of messages stored in log instance
- cleanup in Log code and exported functions names
- tests

### DIFF
--- a/src/js/utils/debug.js
+++ b/src/js/utils/debug.js
@@ -1,9 +1,5 @@
 /* @flow */
 
-// https://stackoverflow.com/questions/7505623/colors-in-javascript-console
-// https://github.com/pimterry/loglevel/blob/master/lib/loglevel.js
-
-// http://www.color-hex.com/color-palette/5016
 const colors: {[k: string]: string} = {
     // green
     'DescriptorStream': 'color: #77ab59',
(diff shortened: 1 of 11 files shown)


src/js/utils/debug.js
```diff
@@ -17,51 +13,56 @@ const colors: {[k: string]: string} = {
...
-    addMessage(level: string, prefix: string, ...args: Array<any>): void {
+    addMessage(level: string, prefix: string, ...args: any[]) {
         this.messages.push({
             level: level,
             prefix: prefix,
             message: args,
             timestamp: new Date().getTime(),
         });
+        if (this.messages.length > MAX_ENTRIES) {
+            this.messages.pop();
```
slowbackspace: Maybe I am missing something, but this way only first <MAX_ENTRIES> will be logged, each message after hitting this limit will be pushed at the end of the array and then immediately popped from it. Shouldn't we log last X messages instead?

szymonlesisz: nice! good catch. i've mindlessly copy/pasted it [from suite](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/reducers/suite/logReducer.ts#L38) :)

szymonlesisz: addressed in suite as well https://github.com/trezor/trezor-suite/pull/3494

slowbackspace (approved):

PR #773, opened 2021-03-02
Fix/blockchain link dispose

szymonlesisz:
- call dispose on blockchain-link instance after error (terminate worker)
- remove unnecessary async/await from Blockchain class


slowbackspace (approved):
much better, see https://github.com/trezor/trezor-suite/issues/3490#issuecomment-789090454

keraf (approved):

PR #777, opened 2021-03-05
Chore/prettier

szymonlesisz:
implementing prettier + eslint with the same rules as trezor-suite* 

_`* actually 3 rules are intentionally disabled since they required more work`_

changes:
- 98% of changes are result of `eslint --fix` (line brakes, destructuring, commas...)
- removed obvious/unnecessary types assignment / return types
- type `Array<X>` to `X[]`
- type `Object` to `any`
- arrow functions `function foo() {}` to `const foo = () => {}`

i was not able to configure prettier + flowtype to enforce semicolon delimiter in type declarations, this is why ALL flowtype types are changed.
flowtype support in prettier is kinda buggy, interface declarations are using `;` while type declarations are using `,`
i guess we can live with that until typescript migration


examples/electron/renderer-with-popup/src/renderer.js
```diff
@@ -1,7 +1,18 @@
 // TrezorConnect is injected as inline script in html
 // therefore it doesn't need to included into node_modules
 // get reference straight from window object
-const TrezorConnect = window.TrezorConnect;
+const { TrezorConnect } = window;
+
+// print log helper
```
tomexx, line 6: we could put it to an helper lib as it is duplicated in other files

scripts/build-npm-extended.js
```diff
@@ -2,42 +2,40 @@ import fse from 'fs-extra';
...
         }
         return true;
     },
 });
 
 // copy typescript
-fse.copySync(ts, `${lib}/typescript`, {
-    filter: function (src, dest) {
+fse.copySync(TS, `${LIB}/typescript`, {
+    filter: (src, _dest) => {
```
tomexx, line 29: `_dest` could be omitted

src/js/core/methods/GetAccountInfo.js
```diff
@@ -125,9 +135,9 @@ export default class GetAccountInfo extends AbstractMethod {
 
             // prepare html for popup
             const str: string[] = [];
-            Object.keys(keys).forEach((k, i, a) => {
+            Object.keys(keys).forEach((k, _i, _a) => {
                 const details = keys[k];
-                details.values.forEach((acc, i) => {
+                details.values.forEach(acc => {
```
tomexx, line 140: let's update vars in commented code as well

src/js/iframe/iframe.js
```diff
@@ -64,7 +65,7 @@ const handleMessage = (event: PostMessageEvent): void => {
                 fail('POPUP.HANDSHAKE: popupMessagePort not found');
                 return;
             }
-            _popupMessagePort = event.ports[0];
+            [_popupMessagePort] = event.ports;
```
tomexx, line 68: doublecheck

tomexx (approved):
I went trough it. I didn't try to understand everything but I have some overview now, thx.
I pointed out minor nitpicks

examples/electron/renderer-with-popup/src/renderer.js
szymonlesisz: this is only example, it should be straightforward and copy-pasteable without unnecessary dependencies, utils, plugins.
only essentials

src/js/core/methods/GetAccountInfo.js
szymonlesisz: which vars do you mean?

src/js/iframe/iframe.js
szymonlesisz: doublecheck of what? this is only destructuring, transpiled by babel to:
`
var _event$ports = event.ports;
_popupMessagePort = _event$ports[0];
`

src/js/core/methods/GetAccountInfo.js
tomexx: That was just a small nitpick - I ment the commented out code below. But when you uncomment the code you will change fn's arguments `_i` to `i` anyway, so it doesn't matter.

tomexx (approved):

keraf (approved):

PR #785, opened 2021-03-23
Fix: safety checks type

goodhoko:
Fixes the type of the `SafetyChecksLevel` enum and documents the process of transforming `.proto` definitions to Flow/TypeScript type definitions.

Exposes `safety_checks` on `ApplySetting` a prerequisite for https://github.com/trezor/trezor-suite/pull/3590/commits/c5b45260832d560bd642d68c017c11809cc15012.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,15 +8,14 @@ stages:
 before_script:
   - nix-shell --run "yarn"
 
-# Build & Lint
-flow lint:
+verify:
   stage: build
   script:
     - nix-shell --run "yarn flow"
+    - nix-shell --run "yarn typescript"
     - nix-shell --run "yarn lint"
     - nix-shell --run "yarn test:unit"
 
-# Build & Lint
 build:
   stage: build
   script:
--- a/docs/protobuf-type-definitions.md
+++ b/docs/protobuf-type-definitions.md
@@ -1,0 +1,48 @@
+# From Protobuf to TypeScript and Flow
+
+This document describes how the definitions of protobuf messages maintained in the
+firmware repo are semi-automatically translated to Flow and TypeScript definitions that are used in Connect codebase and by Connect users respectively.
+
+## The Pipeline
+
+The beginning and source of truth are the `.proto` definitions in the [firmware repository](https://github.com/trezor/trezor-firmware/tree/master/common/protob). These are duplicated as read-only in the [trezor-common](https://github.com/trezor/trezor-common) repository.
+
+`trezor-common` is included in Connect as a git submodule mounted at `submodules/trezor-common`.
+
+Here, `.proto` definitions are translated to a JSON format using [proto2js](https://www.npmjs.com/package/proto2js) package. This JSON is used on runtime by the [trezor-link](https://github.com/trezor/trezor-link/) (de)serialization logic and to generate the Flow and Typescript definitions.
+
+The JSON is transformed to Flow and/or TypeScript definitions by a script in `scripts/protobuf-types.js`. The script also applies 'patches' I.e. after-the-fact fixes manually described in `scripts/protobuf-patches.js`. The patches compensate for/fix
+- The source `.proto` definitions that do not reflect the actual business logic. Usually fields marked as required which are in fact optional.
+- Fields typed as `uint32`, `uint64`, `sint32`, `sint64` in protobuf that need to be represented as strings in runtime because of javascript number's insufficient range. Runtime conversion is handled automatically by `trezor-link`.
+- Similarly, fields typed as `bytes` in protobuf may be represented as hexadecimal `string`, `Buffer`, `Uint8Array` or `Array<number>` in runtime.
+- Optional protobuf fields that get typed as `<T> | undefined` but are in fact deserialized as `<T> | null`. This could be handled globally by `trezor-link`. The patches exist mainly for historical reasons.
+
+All these steps are done manually and all the generated files are tracked in git. It's also not uncommon to circumvent
+some step by eg. generating the messages.json file not from the Common submodule but directly from the firmware repo.
+
+## Example
+
+```bash
+# Clone the connect repo.
+git clone git@github.com:trezor/connect.git
+cd connect
+# Fetch all git submodules.
+make submodules
+# Install the node modules - including the `proto2js` package required in the next step.
+yarn
+# Transform the .proto definitions to JSON.
+# JSON goes to `src/data/messages/messagesN.json`.
+# Note: This make target also immediately runs the flow/TS type generation,
+# but because it implicitly reads from the old JSON it's basically a noop.
+# The flow/TS generation must be run again once the JSON is moved to the expected location.
+make protobuf
+# Observe the diff.
+git diff --no-index -- src/data/messages/messages.json src/data/messages/messagesN.json
+# Replace the old JSON with the new one.
+# In case of breaking changes this workflow is more complicated as it must account for versioning.
+# See the Versions section in ../README.md.
+mv src/data/messages/messagesN.json  src/data/messages/messages.json
+# Generate Flow and TypeScript definitions. (Implicitly reads from `src/data/messages/messages.json`)
+node scripts/protobuf-types.js
+node scripts/protobuf-types.js typescript
+```
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
         "test:unit": "jest --verbose -c jest.config.unit.js",
         "test:integration": "./tests/run.sh",
         "flow": "flow check .",
+        "typescript": "tsc --project src/ts/types/tsconfig.json",
         "lint": "eslint .",
         "lint:fix": "eslint --fix ."
     },
@@ -105,6 +106,7 @@
         "terser-webpack-plugin": "4.2.2",
         "tiny-worker": "^2.3.0",
         "trezor-link": "1.7.3",
+        "typescript": "^4.2.4",
         "uglify-es": "3.3.9",
         "version-bump-prompt": "6.1.0",
         "webpack": "4.44.2",
--- a/src/ts/types/__tests__/management.ts
+++ b/src/ts/types/__tests__/management.ts
@@ -23,6 +23,7 @@ export const management = async () => {
         display_rotation: 180,
         use_passphrase: true,
         label: 'My Trezor',
+        safety_checks: 'Strict',
     });
 
     TrezorConnect.backupDevice({});
--- a/src/ts/types/__tests__/management.ts
+++ b/src/ts/types/__tests__/management.ts
@@ -32,7 +32,7 @@ export const management = async () => {
     });
 
     TrezorConnect.firmwareUpdate({
-        payload: new ArrayBuffer(0),
+        binary: Buffer.from('abcd'),
     });
 
     TrezorConnect.recoveryDevice({
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,6 +39,7 @@ jobs:
         path: node_modules
         key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
     - run: yarn flow
+    - run: yarn typescript
     - run: yarn lint
     - run: yarn test:unit
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Not Yet Released
+
+### Fixed
+- Incorrect typescript type of the `binary` field of the `firmwareUpdate`'s params'.
+
 # 8.1.25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Not Yet Released
 
+### Added
+- `safety_checks` field to the `ApplySettings` interface which lets you set the `safety_checks` Trezor feature.
+
 ### Fixed
 - Incorrect type of the `safety_checks` Trezor feature.
 - Incorrect typescript type of the `binary` field of the `firmwareUpdate`'s params'.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Not Yet Released
 
 ### Fixed
+- Incorrect type of the `safety_checks` Trezor feature.
 - Incorrect typescript type of the `binary` field of the `firmwareUpdate`'s params'.
 
 # 8.1.25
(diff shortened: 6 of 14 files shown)


docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,49 @@
...
+## Example
+
+```bash
+# Clone the connect repo.
+git clone git@github.com:trezor/connect.git
+cd connect
+# Fetch all the submodules.
+# Technically, `--recursive` isn't needed to fetch .proto files.
+# But better be defensive.
+git submodule update --init --recursive
```
szymonlesisz:
```suggestion
make submodules
```

docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,49 @@
+# From Protobuf to TypeScript and Flow
+
+This document describes how the definitions of protobuf messages maintained in the
+firmware repo are semi-automatically translated to Flow and TypeScript definitions that are used in Connect codebase and by Connect users respectively.
+
+## The Pipeline
+
+Firmware team maintains the `.proto` definitions [in the firmware repo](https://github.com/trezor/trezor-firmware/tree/master/common/protob) and duplicates them in the [trezor-common](https://github.com/trezor/trezor-common) repository.
```
szymonlesisz: `.proto` definitions are maintained in the firmware repository and exported in to the trezor-common repository as a read only copy.

docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,49 @@
...
+
+`trezor-common` is included in Connect as a git submodule mounted at `submodules/trezor-common`.
+
+These `.proto` definitions are translated to a JSON format using [proto2js](https://www.npmjs.com/package/proto2js) package. This JSON is used on runtime by the [Link](https://github.com/trezor/trezor-link/) (de)serialization logic and to generate the Flow and Typescript definitions.
+
+The JSON is transformed to Flow and/or TypeScript definitions by a script in `scripts/protobuf-types.js`. The script also applies 'patches' I.e. after-the-fact fixes manually described in `scripts/protobuf-patches.js`. The patches compensate for/fix
+- The source `.proto` definitions that do not reflect the actual business logic. Usually fields marked as required which are in fact optional.
+- Fields that are transferred as numbers in protobuf but need to be represented as strings in JS because JS's number doesn't have sufficient range. Runtime conversion is handled automatically by Link.
```
szymonlesisz:
```suggestion
- Fields expected as `uint32/uint64` and `sint32/sint64` in protobuf while in fact they are represented as strings because of javascript number insufficient range. Runtime conversion is handled automatically by `trezor-link`.
- Fields expected as `bytes` in protobuf  while in fact they are represented as hex strings. Runtime conversion is handled automatically by `trezor-link`.
```

docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,49 @@
...
+- Optional protobuf fields that get typed as `<T> | undefined` but are in fact deserialized as `<T> | null`. This could be handled globally by Link. The patches exist mainly for historical reasons.
```
szymonlesisz:
```suggestion
- Optional protobuf fields that get typed as `<T> | undefined` but are in fact deserialized as `<T> | null`. This could be handled globally by `trezor-link`. The patches exist mainly for historical reasons.
```

docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,49 @@
...
+This document describes how the definitions of protobuf messages maintained in the
+firmware repo are semi-automatically translated to Flow and TypeScript definitions that are used in Connect codebase and by Connect users respectively.
+
+## The Pipeline
+
+Firmware team maintains the `.proto` definitions [in the firmware repo](https://github.com/trezor/trezor-firmware/tree/master/common/protob) and duplicates them in the [trezor-common](https://github.com/trezor/trezor-common) repository.
+
+`trezor-common` is included in Connect as a git submodule mounted at `submodules/trezor-common`.
+
+These `.proto` definitions are translated to a JSON format using [proto2js](https://www.npmjs.com/package/proto2js) package. This JSON is used on runtime by the [Link](https://github.com/trezor/trezor-link/) (de)serialization logic and to generate the Flow and Typescript definitions.
```
szymonlesisz:
```suggestion
`.proto` definitions are translated to a JSON format using [proto2js](https://www.npmjs.com/package/proto2js) package. This JSON is used on runtime by the [trezor-link](https://github.com/trezor/trezor-link/) (de)serialization logic and to generate the Flow and Typescript definitions.
```

szymonlesisz (changes requested):
great! just few nitpicks in docs
good job :)

goodhoko:
@szymonlesisz Thanks.

I incorporated all your remarks. Rebased on master. Fixed the linting.

I also added one more commit that exposes the `safety_checks` field on `ApplySettings`. This enables us to do the safety checks setting. ([PR](https://github.com/trezor/trezor-suite/pull/3590)) There's already a patch in that PR but since this [alters the runtime as well](https://github.com/trezor/connect/pull/785/commits/7131ee6e54f2ea1d9f8e44779f2d5f93877ced90#diff-de17dcd28d2db3b19316e0d62272a76b5fc68b44515c1da91d1a0413473f4b03R29) I guess we need to wait for actual release of connect before releasing suite - a patch doesn't suffice as in suite-web the connect library is pulled from `connect.trezor.io`.

Could you give this one more check, please? .)

docs/protobuf-type-definitions.md
goodhoko: Edited as suggested. Resolving.

docs/protobuf-type-definitions.md
goodhoko: Edited as suggested. Resolving.

docs/protobuf-type-definitions.md
goodhoko: Edited. Pls check. .)

docs/protobuf-type-definitions.md
goodhoko: Edited. Pls check.

docs/protobuf-type-definitions.md
goodhoko: Edited as suggested. Resolving.

docs/protobuf-type-definitions.md
szymonlesisz: once you are using word "repo" and another time "repository".
what about 3rd party contributors that maintain protobuf messages and they are not members of firmware team? (like for example cardano)
that is why i've suggested this change

docs/protobuf-type-definitions.md
szymonlesisz: you didn't mentioned suggested `bytes` as hex string

docs/protobuf-type-definitions.md
goodhoko: > once you are using word "repo" and another time "repository".

Fixed. Thx.

> what about 3rd party contributors that maintain protobuf messages and they are not members of firmware team? 

Good point. It didn't appear to me at first sight (still didn't get used to working in the open) and I wanted the text to be more personal and less passive/technical.

Hopefully fixed now. .)

docs/protobuf-type-definitions.md
goodhoko: > you didn't mentioned ...

You're right. Added that as well.

> ... suggested ...

though I can't see where it was suggested? Maybe I'm missing something? o.O

package.json
```diff
@@ -42,6 +42,7 @@
         "test:unit": "jest --verbose -c jest.config.unit.js",
         "test:integration": "./tests/run.sh",
         "flow": "flow check .",
+        "typescript": "tsc --project src/ts/types/tsconfig.json",
```
szymonlesisz, line 46:
```suggestion
        "type-check": "tsc --project src/ts/types/tsconfig.json",
```
for future consistency with suite scripts

docs/protobuf-type-definitions.md
```diff
@@ -0,0 +1,48 @@
...
+
+`trezor-common` is included in Connect as a git submodule mounted at `submodules/trezor-common`.
+
+Here, `.proto` definitions are translated to a JSON format using [proto2js](https://www.npmjs.com/package/proto2js) package. This JSON is used on runtime by the [trezor-link](https://github.com/trezor/trezor-link/) (de)serialization logic and to generate the Flow and Typescript definitions.
+
+The JSON is transformed to Flow and/or TypeScript definitions by a script in `scripts/protobuf-types.js`. The script also applies 'patches' I.e. after-the-fact fixes manually described in `scripts/protobuf-patches.js`. The patches compensate for/fix
+- The source `.proto` definitions that do not reflect the actual business logic. Usually fields marked as required which are in fact optional.
+- Fields typed as `uint32`, `uint64`, `sint32`, `sint64` in protobuf that need to be represented as strings in runtime because of javascript number's insufficient range. Runtime conversion is handled automatically by `trezor-link`.
+- Similarly, fields typed as `bytes` in protobuf that are represented as base64 string in runtime.
```
szymonlesisz:
```suggestion
- Similarly, fields typed as `bytes` in protobuf may be represented as hexadecimal `string`, `Buffer`, `Uint8Array` or `Array<number>` in runtime.
```

see
https://github.com/trezor/connect/blob/develop/scripts/protobuf-patches.js#L111
or
https://github.com/trezor/connect/blob/develop/scripts/protobuf-patches.js#L133-L134
or
https://github.com/trezor/connect/blob/develop/scripts/protobuf-types.js#L20

szymonlesisz (changes requested):

.gitlab-ci.yml
```diff
@@ -8,15 +8,14 @@ stages:
 before_script:
   - nix-shell --run "yarn"
 
-# Build & Lint
-flow lint:
+verify:
   stage: build
   script:
     - nix-shell --run "yarn flow"
+    - nix-shell --run "yarn typescript"
```
szymonlesisz, line 15: add this job to github actions as well:
https://github.com/trezor/connect/blob/develop/.github/workflows/tests.yml#L41

src/ts/types/__tests__/management.ts
```diff
@@ -32,7 +33,7 @@ export const management = async () => {
     });
 
     TrezorConnect.firmwareUpdate({
-        payload: new ArrayBuffer(0),
+        binary: [],
```
szymonlesisz:
```suggestion
        binary: Buffer.from('abcd'),
```
ts should scream here, if you use array, no?

docs/protobuf-type-definitions.md
goodhoko: Ah, so this is even wider. Thanks for linking the examples!

Suggestion applied [here](https://github.com/trezor/connect/pull/785/commits/e8efb21fe44d752b3e76a151e7eb661a46502087).

package.json
goodhoko: I'm not sure about this. Future consistency is a factor, for sure, but on the other hand I'm more concerned about the local scope here which is that we have two type systems: flow and typescript and it would be good for the script names to be explicit about that.

What about having scoped names like `type-check:flow` and `type-check:typescript`? This goes conceptually well with Suite but reflects the state of Connect.

src/ts/types/__tests__/management.ts
goodhoko: Actually it does not, quite the opposite, applying your suggestion results in error
```
type 'Buffer' is missing the following properties from type 'number[]': pop, push, concat, shift, and 3 more.
36         binary: Buffer.from('abcd'),
           ~~~~~~
```

the `TrezorConnect.firmwareUpdate` [is typed](https://github.com/trezor/connect/blob/e8efb21fe44d752b3e76a151e7eb661a46502087/src/ts/types/api.d.ts#L347-L349) as
```typescript
function firmwareUpdate(
        params: P.CommonParams & (Mgmnt.FirmwareUpdate | Mgmnt.FirmwareUpdateBinary),
    ): P.Response<P.DefaultMessage>;
```

and the `FirmwareUpdate` and `FirmwareUpdateBinary` are defined [here](https://github.com/trezor/connect/blob/e8efb21fe44d752b3e76a151e7eb661a46502087/src/ts/types/trezor/management.d.ts#L31-L39) and indeed the `binary` field is of type `number[]`.

I didn't consider that as wrong and so I simply alligned the type here in the "tests" rather than in the API which would be a breaking change. But maybe I'm missing something...?

package.json
szymonlesisz: ok, leave it as typescript. it's a detail which could be easily changed

src/ts/types/__tests__/management.ts
szymonlesisz: actually this is a bug in ts definitions.
as you noticed ts is written by hand and never tested before (and it's not used in compile time) 
flowtype and protobuf definitions (and of course runtime code) are the source of truth here:
https://github.com/trezor/connect/blob/develop/src/js/types/trezor/protobuf.js#L567
https://github.com/trezor/connect/blob/develop/src/js/types/trezor/management.js#L31
https://github.com/trezor/connect/blob/develop/src/js/core/methods/FirmwareUpdate.js#L38

this `FirmwareUpdateBinary` was prepared for a feature called "upload custom firmware" in trezor-suite and never implemented (same as safety_checks) this is the reason for not catching this bug before.

don't worry. this is not a breaking change, it's a bug fix

goodhoko: Aaah, ok. I changed the type to `Buffer` in https://github.com/trezor/connect/pull/785/commits/f513e33a7094ef1d6a400af511e732e3857130ee

I also added a record to changelog.

.gitlab-ci.yml
goodhoko: added in https://github.com/trezor/connect/pull/785/commits/f513e33a7094ef1d6a400af511e732e3857130ee

goodhoko:
@szymonlesisz Ok I adressed both open threads and added records to changelog. (All done withing fixup commits.)

The failing CI doesn't seem to be related to this PR's changelist. (Fails on unavailable blockchain_link backends.) What can I do about it? Maybe a rebase on master would help? ...I didn't want to force push.

The changes of this PR needs to be released to unblock the [dependent PR](https://github.com/trezor/trezor-suite/pull/3590) in Suite because we [touched the runtime](https://github.com/trezor/connect/pull/785/commits/54f80c461ea132c3ee23cbe17f8054772c1f630e#diff-de17dcd28d2db3b19316e0d62272a76b5fc68b44515c1da91d1a0413473f4b03R29) here.

What are the next steps towards a release?

szymonlesisz (approved):
great, you can squash it now
dont worry about failing GA tests, gitlab ci is passing

goodhoko:
Squashed and rebased on `develop`. This is good from my side. Feel free to merge.

PR #786, opened 2021-03-24
fix: rbf opreturn tx

szymonlesisz:
decode op_return_data while transforming original transaction (rbf case only)

added test with RBF + OPRETURN combination

### DIFF
--- a/src/js/core/methods/tx/refTx.js
+++ b/src/js/core/methods/tx/refTx.js
@@ -1,6 +1,10 @@
 /* @flow */
 
-import { coins as BitcoinJSCoins, Transaction as BitcoinJsTransaction } from '@trezor/utxo-lib';
+import {
+    coins as BitcoinJSCoins,
+    script as BitcoinJSScript,
+    Transaction as BitcoinJsTransaction,
+} from '@trezor/utxo-lib';
 import type { TypedRawTransaction } from '@trezor/blockchain-link';
 import type { Input as BitcoinJsInput, Output as BitcoinJsOutput } from '@trezor/utxo-lib';
 // local modules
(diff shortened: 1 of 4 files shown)


src/js/core/methods/tx/refTx.js
```diff
@@ -102,6 +102,13 @@ export const transformOrigTransactions = (
 
         // outputs, required by TXORIGOUTPUT (TxAckOutput) request from Trezor
         const outputsMap = (output: BitcoinJsOutput, i: number) => {
+            if (!vout[i].isAddress && vout[i].hex) {
+                return {
+                    script_type: 'PAYTOOPRETURN',
+                    amount: '0',
+                    op_return_data: vout[i].hex.substring(4), // strip 2 bytes: 6aXX. blockbook sends it as scriptPubKey. where 6a - OP_RETURN and XX - data len
```
andrewkozlik: This will work only if the `op_return_data` is 75 bytes in length or less, because the second byte in the scriptPubKey is a data-push operation, see https://en.bitcoin.it/wiki/Script#Constants.

I would also recommend checking that `vout[i].hex[0:2] == "6a"`, parsing the data-push to get the length and checking that there is no trailing data. If these conditions are not satisfied then throw "Unsupported script type" to ensure that errors are caught in the right place.

andrewkozlik: Example of an extra-long piece of op_return_data: https://btc1.trezor.io/tx/d29c9c0e8e4d2a9790922af73f0b8d51f0bd4bb19940d9cf910ead8fbe85bc9b

szymonlesisz: fortunately we already have a code for decoding data :) this should throw an error in case of invalid script type
thanks for pointing this up! appreciate it!

andrewkozlik: LGTM

keraf (approved):

PR #788, opened 2021-03-29
feat: update ltc and tltc fee options

mcudev:
this does a couple things.

first, this sets a blockbook node for litecoin testnet. **this assumes that trezor would run a litecoin testnet blockbook node**. if not, i'd probably suggest using `https://tltc1.test` as that is one of the special use domains unless (someone else wants to setup and run a blockbook node). i've been using `tltc1.test` with a mapping in my `/etc/hosts` file while testing trezor-suite changes.

second, this configures the fee options to what both networks currently support. i've been able to successfully send 1 sat/vbyte transactions on both networks with no problem.

### DIFF
--- a/src/js/core/methods/tx/Fees.js
+++ b/src/js/core/methods/tx/Fees.js
@@ -16,6 +16,20 @@ const BLOCKS = {
         economy: 36,
         low: 144,
     },
+    ltc: {
+        // blocktime ~ 150sec.
+        high: 1,
+        normal: 24,
+        economy: 144,
+        low: 576,
+    },
+    tltc: {
+        // blocktime ~ 150sec.
+        high: 1,
+        normal: 24,
+        economy: 144,
+        low: 576,
+    },
     bch: {
         // blocktime ~ 600sec.
         high: 1,


szymonlesisz:
Hi @mcudev, 
the file you've changed is auto generated from [trezor-firmware](https://github.com/trezor/trezor-firmware/blob/master/common/defs/bitcoin/litecoin_testnet.json) definitions.

regarding running litecoin testnet node, i dont think there are any plans for that.
if you are running your own instance of blockbook for `tLTC` you can use it in connect by setting up custom backend:

```
TrezorConnect.blockchainSetCustomBackend({
    coin: 'tltc',
    blockchainLink: {
        type: 'blockbook',
        url: ['your-custom-url'],
    },
});
```


what is the motivation for adding multiple fee levels with the same values?

mcudev:
@szymonlesisz i knew that i had seen those coin configs in trezor-firmware. but, i didn't understand how they got here until your comment spurred me to dig into it. now i think i understand, thanks! i probably made this same mistake in a few other commits that i pushed up today for other PR's, ugh, lol. i'll push a different commit with only the changs to `src/js/core/methods/tx/Fees.js`

i was hoping that i could talk you guys into hosting a tltc blockbook node so that i could do some trezor-suite testing easier (for example, https://github.com/trezor/trezor-suite/pull/3574) and new user introductions (https://github.com/trezor/trezor-suite/issues/3568). i understand if you guys don't want to though -- all the overhead adds up.

these changes are part of getting trezor-suite refreshed for litecoin and getting litecoin testnet working similarly. i figured that i'd add all the fee levels that bitcoin has. these are the defaults and then the fee estimates come from the blockbook nodes and can vary if activity on the networks dictate. these aren't critical changes. i just know that the normal default fee of 1000 was wrong for LTC, so i figured while i was making the change, that i'd go all the way.

mcudev:
ok, i updated this so that it goes along logically with https://github.com/trezor/trezor-firmware/pull/1555

mcudev:
changed my mind on wanting/requesting this

PR #792, opened 2021-03-31
feat: configure dev webpack server https certificate with basicConstraints CA

mcudev:
This scratches one of my own itches. In short, this sets a self-signed cert for the dev connect server that has the basicConstraints CAset to true so that Chrome will allow importing.

More detailed explanation:
I've had to make some connect changes and test them locally.
These changes are used via my local trezor-suite server.
So, I point my trezor-suite at my localdev connect server and then access the suite in my fresh browser instance.
Of course, the the browser blocks the request because the cert of the connect server is not trusted.
I can manually open https://localhost:8088/ in another tab and accept the security warning. But, I have to do this each time I kill the browser and start again.
Normally, I'd import the connect server's cert into Chrome and be done with it. But, the cert does not have the CA basic constraint set to true and thus the current Chrome version blocks that action.

The cert was generated with:

`openssl req -x509 -newkey rsa:4096 -sha256 -days 3650 -nodes -keyout connect_dev.key -out connect_dev.crt -subj '/O=INSECURE AUTHORITY/OU=DEVELOPMENT/CN=localhost' -extensions san -config <( echo '[req]'; echo 'distinguished_name=req'; echo '[san]'; echo 'subjectAltName=DNS:localhost'; echo 'basicConstraints=critical,CA:true,pathlen:0';)`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ coverage
 
 # Cache
 __pycache__
+
+# dev
+webpack/connect_dev.crt
+webpack/connect_dev.key
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
         "./lib/env/node/networkUtils": "./lib/env/react-native/networkUtils"
     },
     "scripts": {
+        "predev": "node webpack/generate_dev_cert.js",
         "dev": "webpack-dev-server --config ./webpack/config.dev.babel.js --mode development",
         "build": "rm -rf build && webpack --config ./webpack/config.prod.babel.js --progress",
         "build:node": "rm -rf build && webpack --config ./webpack/config.nodejs.babel.js --progress",
--- a/webpack/config.dev.babel.js
+++ b/webpack/config.dev.babel.js
@@ -1,7 +1,7 @@
 import webpack from 'webpack';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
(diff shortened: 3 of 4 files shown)


keraf (approved):
Change looks good to me, thanks for the contribution! This will definitely make it a bit more smoother to work with connect :slightly_smiling_face:  

The only suggestion I would have would be to add the key gen command (with the correct paths) to the package.json scripts.

mcudev:
> The only suggestion I would have would be to add the key gen command (with the correct paths) to the package.json scripts.

Interesting idea. I suppose that I could .gitignore the cert and the key and then in the package.json script, check for presence of the files and generate them if not present. It would add a dependency on openssl being installed (which might exist implicitly already, I don't know, just thinking as I type). It would probably lead to a few extra cert imports over time. The benefit is that we don't have the same cert trusted by a group of developers/targets.

Is that benefit worth the change in approach? Maybe. I don't have a strong opinion. If it's worth it, I'll tinker with setting it up a little later today or tomorrow.

mcudev:
@keraf I just force pushed an update that generates the cert at predev in the npm scripts. I added a script, `webpack/generate_dev_cert.js` that depends on the existing `selfsigned` dependency rather than adding a command line dependency on openssl. I think this accomplishes what you mentioned.

mcudev:
I just realized that the linter was complaining at me so I ran the fixer and repushed it. Should be good now.

szymonlesisz (approved):

szymonlesisz:
thanks @mcudev i was thinking about it for a long time (never had time to do this tho)

PR #780, opened 2021-03-11
Chore/webpack5

szymonlesisz:
### update webpack@5
manually added nodejs polyfills (crypto, assert...)

### karma tests on production build
fixtures are shared with integration tests

Issue #781, opened 2021-03-18
CI / tests tracking issue


szymonlesisz:
- [ ] OS runners matrix (osx + windows)
- [ ] collect coverage (jest, karma) from separated CI jobs
- [ ] karma: firefox
- [ ] karma: handle new window opened (popup mode)
- [ ] karma: move jasmine matcher (toMatchObject, expect.any in getFeatures fixtures) move to karma plugin
- [ ] karma: serve from https (webusb case)
- [ ] unit tests for method params
- [ ] unify tests with https://github.com/trezor/trezor-firmware/tree/master/common/tests/fixtures/cardano

PR #798, opened 2021-04-16
tests with cache

szymonlesisz:
- added referenced transactions and account info cache to reduce network traffic (and fixing github actions tests)
- improve `tests/run` script and test docs
- added runtime referenced transaction validation
- type fixes
- removed legacy tests, karma tests will be restored here: https://github.com/trezor/connect/pull/780


tests/README.md
```diff
@@ -1,28 +1,62 @@
-# Connect integration tests
+# trezor-connect tests
+## Continuous Integration
+Connect tests are running in two environments:
+- [SatoshiLabs GitLab CI](https://gitlab.com/satoshilabs/trezor/connect/-/pipelines) NixOS runner (trezor-user-env)
+- [Github Actions CI](https://github.com/trezor/connect/actions) Ubuntu runner
```
tsusanka: Maybe we could elaborate that in Gitlab we run directly inside trezor-user-env while GH Actions uses the image via `/.run.sh` therefore kind of "mimicks" when run locally?

szymonlesisz: 7255f9f

tests/run.sh
```diff
@@ -1,93 +1,106 @@
...
-set -e
-
-function cleanup() {
-  echo "Cleaning up"
-  echo "Stopping container with an ID $id"
-  # uncomment to show logs from container, useful for debugging
-  # docker logs $id
-  docker stop "$id" && echo "Stopped"
+# This script runs trezor-connect tests.
+# It spin up trezor-user-env and set required evironment variables.
```
goodhoko:
```suggestion
# It spins up trezor-user-env and sets required evironment variables.
```

tests/run.sh
```diff
@@ -1,93 +1,106 @@
...
-  echo "  -f       Use specific firmware version, for example: 2.1.4., 2.3.0"
-  echo "  -i       Included methods only, for example: applySettings,signTransaction"
-  echo "  -e       All methods except excluded, for example: applySettings,signTransaction"
-  echo "  -c       Collect coverage"
+  echo "  -d       Disable docker. Useful when running own instance of trezor-user-env. default: enabled"
+  echo "  -c       Disable backend cache. default: enabled"
+  echo "  -f       Use specific firmware version, example: 2.1.4, 1.8.0 default: 2-master"
+  echo "  -i       Included methods only, example: applySettings,signTransaction"
+  echo "  -e       All methods except excluded, example: applySettings,signTransaction"
+  echo "  -s       actual test script. default: 'yarn test:integration'"
```
goodhoko: nitpick: I really like if the options are ordered alphabetically it's then easy to find what you're looking for at a glance.
```suggestion
  echo "  -c       Disable backend cache. default: enabled"
  echo "  -d       Disable docker. Useful when running own instance of trezor-user-env. default: enabled"
  echo "  -e       All methods except excluded, example: applySettings,signTransaction"
  echo "  -f       Use specific firmware version, example: 2.1.4, 1.8.0 default: 2-master"
  echo "  -i       Included methods only, example: applySettings,signTransaction"
  echo "  -s       actual test script. default: 'yarn test:integration'"
```

src/js/core/methods/tx/refTx.js
```diff
@@ -202,3 +204,88 @@ export const transformReferencedTransactions = (
...
+
+// Validate referenced transactions provided by the user.
+// Data sent as response to TxAck needs to be strict.
+// They should not contain any fields unknown/unexpected by protobuf.
+export const validateReferencedTransactions = (
+    txs: RefTransaction[] | typeof undefined,
+    inputs: TxInputType[],
+    outputs: TxOutputType[],
+) => {
+    if (!Array.isArray(txs) || !txs.length) return; // allow empty, they will be downloaded later...
```
goodhoko: I don't know what is our convention but I generally preffer explicit booleans in conditionals. RBY.

```suggestion
    if (!Array.isArray(txs) || txs.length === 0) return; // allow empty, they will be downloaded later...
```

src/js/core/methods/tx/refTx.js
```diff
@@ -202,3 +204,88 @@ export const transformReferencedTransactions = (
...
+                amount: output.amount,
+                script_pubkey: output.script_pubkey,
+                decred_script_version: output.decred_script_version,
+            })),
+        };
+    });
+
+    // check if all required transactions defined by inputs/outputs were provided
+    refTxs.concat(origTxs).forEach(hash => {
+        if (!transformedTxs.find(tx => tx.hash === hash)) {
```
goodhoko, line 285: Similarly here. RBY.

```suggestion
        if (transformedTxs.find(tx => tx.hash === hash) === undefined) {
```

tests/README.md
```diff
@@ -43,3 +43,23 @@ _Note: You can run `trezor-user-env` as standalone process with graphical output
 1. Create or modify file in `tests/__fixtures__`
 1. Make sure it is imported in `tests/__fixtures__/index.js`
 1. Make sure the method you are testing is listed in `.github/workflows/*.yml` and `.gitlab-ci.yml`
+
+## Transactions cache
+Bitcoin-like coins `signTransaction` method require additional data about transactions referenced from used inputs.
+
+Those data are automatically downloaded using backed defined in `coins.json` by default if `refTxs` param is not specified.
```
goodhoko: I guess...?

```suggestion
Those data are automatically downloaded from backend defined in `coins.json` by default if `refTxs` param is not specified.
```

tests/__wscache__/index.js
```diff
@@ -0,0 +1,35 @@
...
+            } catch (error) {
+                console.error(`WS_CACHE parsing error: ${filePath}`);
+                throw error;
+            }
+        }
+    });
+};
+
+// read cache directory
+cacheFiles(path.resolve(__dirname));
```
goodhoko: We export an empty global variable that is later populated. I'd much prefer if the `cacheFiles` function *returned* the results instead of mutating global var (Which is `const`!) and we'd just
```js
export const WS_CACHE = cacheFiles(path.resolve(__dirname));
```

What do you think?

tests/README.md
```diff
@@ -63,3 +63,8 @@ Cached transactions are stored in `tests/__txcache__` directory in the same stru
 Cached transactions are provided to test fixtures via [TX_CACHE](./__txcache__/index.js) utility.
 
 Missing tx json? use [this tool](./__txcache__/gen-reftx.js) to generate it.
+
+## Websocket cache
```
goodhoko, line 73: If I understood this right the Transaction and Websocket caches are not caches in the usual sense of eg. networking caches. I.e. they do not act as transparent layer that can forward requests to the 'origin' if the response isn't available locally. Instead they're more like mocking tools that let you supply your own responses. Right?

If this is the case can we maybe not call these "Caches"? When I read "cache" I imagine something that merely *caches* responses to reduce bandwith but definitely not something that mocks the responses.

goodhoko (approved):
Went thru this quickly. Just a few small remarks. Generally LGTM.

tests/README.md
szymonlesisz: not quite, Transaction cache is called "cache" because i want to keep naming convention with [trezor-firmware](https://github.com/trezor/trezor-firmware/tree/master/tests/txcache). One day those files will be shared via [trezor-common](https://github.com/trezor/connect/issues/781)

Websocket cache are mocked (cached 1:1) responses of `blockchain-link` to reduce unnecessary websocket connection. Maybe we could create websocket server which will provide those cached values normally (using not-mocked blockchain-link) but i think it's not worth the effort and it's out of scope for this PR.

In both cases tests with disabled cache (on gitlab CI) will work with exact same values, but dowloaded from the internet

tsusanka: I agree _cache_ might not be the greatest naming here (but mock isn't either) but as @szymonlesisz puts it I would keep it to have it the same with FW for now.

src/js/core/methods/tx/refTx.js
szymonlesisz: we are using this syntax everywhere, here and in suite. leaving as is

tests/README.md
szymonlesisz: 317be40

src/js/core/methods/tx/refTx.js
szymonlesisz: c6b5906

tests/run.sh
szymonlesisz: 468c833

tests/run.sh
szymonlesisz: 468c833

tests/__wscache__/index.js
szymonlesisz: doesn't really matter since this script will execute before tests, but i've changed it for both 46d2f03 and d3599f8

tests/run.sh
goodhoko: Thanks

tests/run.sh
goodhoko: 🦊

src/js/core/methods/tx/refTx.js
goodhoko: 🦊

tests/README.md
goodhoko: The patch didn't fix the typo in `backed` (should be `backend` I believe).

Not important at all -- feel free to reslove this.

tests/__wscache__/index.js
goodhoko: Thanks!

tests/README.md
goodhoko: > Transaction cache is called "cache" because i want to keep naming convention with trezor-firmware. One day those files will be shared via trezor-common

Allright, didn't know of this parity.

> but i think it's not worth the effort and it's out of scope for this PR.

Agree.

Resolving.

goodhoko:
@szymonlesisz Thanks for addressing all my remarks. This is good from my side (except one totally minor typo).

szymonlesisz:
all fixed and rebased

tests/README.md
szymonlesisz: ah, spell checker didn't catch this :) `backed` is also a word :)
thanks

tests/README.md
szymonlesisz: i guess i can resolve this

tests/device/methods.test.js
```diff
@@ -55,6 +56,10 @@ fixtures.forEach(testCase => {
                     return;
                 }
 
+                // print current test case, `jest` default reporter doesn't log this. see https://github.com/facebook/jest/issues/4471
+                const log = chalk.black.bgYellow.bold(` ${testCase.method}: ${t.description} `);
```
tomasklim:
```suggestion
                const log = `${chalk.black.bgYellow.bold(` ${testCase.method}: `)} ${chalk.bold(
                    t.description,
                )}`;
```

tomasklim: Screenshot 2021-04-26 at 12 17 44

tomasklim (approved):

tests/device/methods.test.js
szymonlesisz: fixed, rebased

PR #803, opened 2021-04-27
Feat/expose protobuf enums

szymonlesisz:
expose protobuf enums/objects to public as [requested](https://github.com/trezor/connect/pull/794#issuecomment-825108329)

### DIFF
--- a/src/js/constants/cardano.js
+++ b/src/js/constants/cardano.js
@@ -10,6 +10,9 @@ export const NETWORK_IDS = Object.freeze({
     testnet: 0,
 });
 
+// constants below are deprecated
+// use `CardanoAddressType`, `CardanoCertificateType` and `CardanoPoolRelayType` from protobuf instead
+
 export const ADDRESS_TYPE = Object.freeze({
     Base: 0,
     Pointer: 4,
--- a/src/js/core/methods/helpers/cardanoCertificate.js
+++ b/src/js/core/methods/helpers/cardanoCertificate.js
@@ -1,7 +1,10 @@
 /* @flow */
 
 import { validateParams } from './paramsValidator';
-import { CERTIFICATE_TYPE, POOL_RELAY_TYPE } from '../../../constants/cardano';
+import {
+    Enum_CardanoCertificateType as CardanoCertificateType,
+    Enum_CardanoPoolRelayType as CardanoPoolRelayType,
+} from '../../../types/trezor/protobuf';
 import { validatePath } from '../../../utils/pathUtils';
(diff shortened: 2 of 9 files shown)


src/js/constants/cardano.js
```diff
@@ -9,24 +9,3 @@ export const NETWORK_IDS = Object.freeze({
     mainnet: 1,
     testnet: 0,
 });
-
-export const ADDRESS_TYPE = Object.freeze({
```
gabrielKerekes, line 13: I would rather leave the constants just to be sure we don't break anything. Maybe we can deprecate them - probably just by mentioning in the docs that these constants have been replaced by the enums and might be replaced in the future. I could add that in #794 after rebasing on top of this. Or is there a better way we could deprecate them?

gabrielKerekes (approved):
LGTM overall! I would maybe just prefer aliasing all the `Enum_Xxx` as `Xxx` e.g. `Enum_CardanoCertificateType as CardanoCertificateType`.

src/js/constants/cardano.js
szymonlesisz: constants restored with comment (there is no mention about it in docs, feel free to update it)
enums aliased

95d8927a3f9d8c5a66eca0d9cd04126cb13eb51f

szymonlesisz: additionally added same aliases in tests 6ed4de2a39f97ff65e0a99931087dc23c2de7a20

src/js/core/methods/helpers/cardanoCertificate.js
```diff
@@ -58,12 +61,12 @@ const validatePoolRelay = (relay: CardanoPoolRelay) => {
...
             );
         }
-    } else if (relay.type === POOL_RELAY_TYPE.SingleHostName) {
+    } else if (relay.type === CardanoPoolRelayType.SINGLE_HOST_NAME) {
         validateParams(relay, [
             { name: 'hostName', type: 'string', obligatory: true },
             { name: 'port', type: 'number', obligatory: true },
         ]);
-    } else if (POOL_RELAY_TYPE.MultipleHostName) {
+    } else if (relay.type === CardanoPoolRelayType.MULTIPLE_HOST_NAME) {
```
szymonlesisz, line 69: @gabrielKerekes whats your opinion on this change?
previous condition would always return true (MULTIPLE_HOST_NAME = 2)

gabrielKerekes: This definitely seems to be a bug - not a serious one, but still a bug. The change makes sense :+1:

PR #808, opened 2021-05-05
Add 1.10.0 firmware

tsusanka:
Just a copy & paste from https://github.com/trezor/webwallet-data/commit/f8ed15a8999689e7692b8fc4c00b7aaef25d8011 . Right?


tsusanka:
So, this is not needed I guess? Since I did it in https://github.com/trezor/webwallet-data/commit/f8ed15a8999689e7692b8fc4c00b7aaef25d8011

PR #818, opened 2021-05-25
fix: device features definitions

tomasklim:
The protobuf definition of device features declares the majority of attributes as optional. However, it is not true, as we are getting all of these optional attributes with `null` value.


goodhoko (approved):
Oh yeah, wanted to do this too before, but then only changed the `safety_checks` field which I needed. It's good to have all the fields right.

szymonlesisz (approved):

PR #819, opened 2021-05-25
fix: revision inconsistency between T1 and T2

tomasklim:
Device features revision fix related to [trezor-suite #3774](https://github.com/trezor/trezor-suite/pull/3774) and [trezor-firmware](https://github.com/trezor/trezor-firmware/issues/1620).

The problem is that T2 uses a different encoding of revision than T1.

Examples of revisions:
T2: 6466303936336563, 623139636266363763 (correct revision encoding from fw point of view)
T1: df0963ec48f01f3d07ffca556e21ff0070cab099, f4424ece1ccb7fc0d6cad00ff840fac287a34f07 (correct revision encoding we want to distribute from trezor-connect to apps to avoid unnecessary encoding)

Useful tools:
Hex to utf8: https://onlineutf8tools.com/convert-hexadecimal-to-utf8
utf8 to hex: https://onlineutf8tools.com/convert-utf8-to-hexadecimal
Regex tester: https://regex101.com/

Before:
https://github.com/trezor/trezor-firmware/commit/6466303936336563

After:
https://github.com/trezor/trezor-firmware/commit/df0963ec
Screenshot 2021-05-25 at 12 47 04


tomasklim:
You can review the pull request. It is based on #818 as it requires correct definition types for CI job.

src/js/utils/__tests__/deviceFeaturesUtils.test.js
```diff
@@ -176,4 +180,37 @@ describe('utils/deviceFeaturesUtils', () => {
...
+
+        // T2
+        expect(parseRevision(featT2)).toEqual(parsedRevisionT2);
+
+        // bootloader
+        expect(parseRevision(featBootloader)).toEqual(parsedRevisionBootloader);
+
+        // UTF-8 T1 revision with only numbers
+        expect(parseRevision(featRevisionOnlyNumbers)).toEqual(parsedRevisionOnlyNumbers);
+    });
```
goodhoko: It seems to me that we're testing 3 scenarios here each with distinct and independent setup and evaluation phases. The test name is also quite ambiguous. Could we split this to 3 separate tests each with a descriptive name? Something like
```suggestion
    it('parses T1 revision', () => {
        const features = {
            revision: 'f4424ece1ccb7fc0d6cad00ff840fac287a34f07',
        };
        const result = 'f4424ece1ccb7fc0d6cad00ff840fac287a34f07';
        
        expect(parseRevision(features)).toEqual(result);
    })
    
    it('parses T2 revision', () => {
        ...
    });
    
    it('null is passed through', () => {
        ...
    });
    
    it('parses T1 revision containing only numbers', () => {
        ...
    });
```

When such test fail it will be much clearer what went wrong on the first sight. .)

src/js/utils/deviceFeaturesUtils.js
```diff
@@ -32,10 +32,12 @@ export const parseCapabilities = (features?: Features): Capability[] => {
     // needs to be "any" since Features.capabilities are declared as string[] but in fact it's a number[]
     const filter = (c: any) => (CAPABILITIES[c - 1] ? [CAPABILITIES[c - 1]] : []);
     // fallback for older firmware
-    if (!features.capabilities || !features.capabilities.length)
+    if (!features.capabilities || !features.capabilities.length) {
         return features.major_version === 1
             ? DEFAULT_CAPABILITIES_T1.flatMap(filter)
             : DEFAULT_CAPABILITIES_TT.flatMap(filter);
+    }
+
```
goodhoko: nit/optional/food for though: This seems like a pure fomating refactor. I'm up for refactoring the code-base piece by piece as we work on it and I support this particular change as well. Let's maybe just think of separating the refactoring from feature implementation by commit each separately. That would have two benefits that go hand-in-hand:
-  easier review: I can put all my focus on the crucial feature-implementing changes and quickly skim over the formatting refactoring commits.
- more refactoring: if we commit the refactoring separately we don't have to worry about it cluttering the new code. We can then boldly refactor whole files in a PR knowing the reviewer can still fully focus on the new code.

src/js/utils/deviceFeaturesUtils.js
```diff
@@ -106,3 +108,33 @@ export const getUnavailableCapabilities = (
     });
     return list;
 };
+
+/**
+ * In current versions of FW, TT uses hex raw bytes and T1 uses UTF-8 representation of revision.
+ * Examples:
+ * T1: df0963ec48f01f3d07ffca556e21ff0070cab099, f4424ece1ccb7fc0d6cad00ff840fac287a34f07
+ * TT: 6466303936336563, 623139636266363763
```
goodhoko:
```suggestion
 * TT: 6466303936336563, 623139636266363763
 * T1: df0963ec48f01f3d07ffca556e21ff0070cab099, f4424ece1ccb7fc0d6cad00ff840fac287a34f07
```

let's preserve relative ordering from the previous sentence. It makes the reading a bit easier.

src/js/utils/deviceFeaturesUtils.js
```diff
@@ -106,3 +108,33 @@ export const getUnavailableCapabilities = (
...
+ * @returns revision in UTF-8 format or null
+ */
+export const parseRevision = (features: Features) => {
+    const { revision } = features;
+
+    // if device is in bootloader mode, revision is null
+    if (!revision) return null;
+
+    // if revision contains at least one a-f character and the rest are numbers, it is in UTF-8 (no encoding needed)
+    if (/^(?=.*[a-fA-f])([a-fA-F0-9]*)$/.test(revision)) {
```
goodhoko: I though that a mere letter is enough to be sure the revision is in the ut8 format. Couldn't we just test for a presence of a letter?

```suggestion
    if (/[a-fA-f]/.test(revision)) {
```

goodhoko: optional: we could consider using the `i` [regex flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags) to make the patterns case insensitive instead of repepating `[A-F]` several times. .)

src/js/utils/__tests__/deviceFeaturesUtils.test.js
tomasklim: It's true that I did sloppy work here. I was inspired by the other tests in this file. I'll fix it.

src/js/utils/deviceFeaturesUtils.js
tomasklim: I will remove it

src/js/utils/__tests__/deviceFeaturesUtils.test.js
goodhoko: Btw the tests can be basically one-liners if we omit the setup expressions and inline them into the `expect` invocations. .)

```js
it('parses T1 revision', () => {
    expect(parseRevision({revision: 'f4424ece1ccb7fc0d6cad00ff840fac287a34f07'))
        .toEqual('f4424ece1ccb7fc0d6cad00ff840fac287a34f07');
})

it('null is passed through', () => {
    expect(parseRevision({revision: null)).toBeNull();
})
```

src/js/utils/deviceFeaturesUtils.js
tomasklim: I need to test that it consists only from numbers and a-f and it contains at least one a-f.

src/js/utils/deviceFeaturesUtils.js
```diff
@@ -106,3 +106,30 @@ export const getUnavailableCapabilities = (
     });
     return list;
 };
+
+/**
+ * There is an inconsistency in representation of device feature revision attribute.
+ * - T1 uses UTF-8 representation. (df0963ec48f01f3d07ffca556e21ff0070cab099)
+ * - T2 uses hex raw bytes representation. (6466303936336563)
+ * To avoid being model specific, in case the inconsistency is fixed, it is required to reliably detect what encoding is used.
```
goodhoko, line 114: I'm really struggling to understand all this encoding marshalling. Please correct me if I'm wrong.

Git identifies revisions by their SHA-1 hash expressed with the standard *[hexadecimal](https://www.wikiwand.com/en/Hexadecimal)* notation -- for example `5ef3efa42638d618e024c3a8f9197ffb03e243b8` for [this commit](https://github.com/trezor/connect/commit/5ef3efa42638d618e024c3a8f9197ffb03e243b8). And that's what we want to ultimately get here.

When a revisions is obtained from a TT device we get the hexadecimal representation of the hash as JS string directly.

When a revisions is obtained from T1 device, however, what we get isn't this hexadecimal representation but a hexadecimal representation of the bytes that form an utf8 sequence containing the original hexadecimal representation.

What we want here is to automatically distinguish these two cases and convert the kinda double-encoded variant to the simple one. Because commit sha only contains alfanumeric characters its utf8 representation encoded as hexadecimal contains only numbers[1]. So we first look if the incoming string contains a letter. If it does, it must be the simple variant and we can return it.

If the incoming string consists of numbers only it is very likely[2] the double-encoded variant hence we decode it (`Buffer.from(<incoming>` 'hex')`) and then put it to an utf8 string (`.toString('utf8')`) which now contains the simple *hexadecimal representation of the hash*.

Before we return this decoded variant we also check that it decoded to a valid hexadecimal. If it did, we return it. If it didn't it might mean that the incoming string was in fact the simple variant or that the data was corrupted.

There's still a small chance that a simple variant would be misclassified as double-encoded if 1) it consisted of numbers only 2) those numbers, read as a hexadecimal, would encode a valid utf8 sequence which would consist of alfa-numeric characters only. I won't try to calculate probability of that. .)

---

The comments in this PR are right, but also confusing because on first sight, they're contradictory. They give `df0963ec48f01f3d07ffca556e21ff0070cab099` as example of "utf8" encoding in opposition to `6466303936336563` a "[raw] hex" encoding. However most people, when shown `df0963ec48f01f3d07ffca556e21ff0070cab099`, would say it's a hexadecimal. I'm not really sure how to make them clearer. ./

Maybe it's just me. But it took me quite big effort to understand what is going on here.


[1] `0-9` maps to `30-39`, `a-f` maps to `61-66` and `A-Z` to `41-46`.
[2] Probability of getting all-numbers sha-1 hash is `(10/16)^40` or roughly `6.8e-9`.

tomasklim: Yes, this is exactly what it means 👍  However, I have difficulties with how to express it clearly by only few words. I will give it another chance tomorrow.

tsusanka: We could also link https://github.com/trezor/trezor-firmware/issues/1620 for more details.

PR #824, opened 2021-05-31
fix: device features unlocked type

tomasklim:
Fixing forgotten change of unlocked attribute in device features.

`unlocked` can be either `boolean` or `null`


goodhoko:
When exactly can it be `null`? I guess when the device doesn't support pin locking? .)

goodhoko (approved):

szymonlesisz (approved):

PR #833, opened 2021-06-22
Fix/unreadable device

szymonlesisz:
Distinguish unreadable device from unacquired device.
Unreadable device error (reason) is passed in unreadable Device.id field.

Device selection is required if unreadable device is present (popup mode)

### DIFF
--- a/src/js/env/browser/browserUtils.js
+++ b/src/js/env/browser/browserUtils.js
@@ -52,6 +52,12 @@ export const getBrowserState = (supportedBrowsers: {
     };
 };
 
+export const getOS = () => {
+    if (!navigator || !navigator.userAgent) return 'unknown';
+    const browser = Bowser.getParser(navigator.userAgent);
+    return browser.getOS().name.toLowerCase();
+};
+
 const getSuggestedPlatform = () => {
     if (!navigator || !navigator.userAgent) return;
     // Find preferred platform using bowser and userAgent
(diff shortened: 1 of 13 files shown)


src/js/env/browser/browserUtils.js
```diff
@@ -51,6 +51,12 @@ export const getBrowserState = (supportedBrowsers: {
     };
 };
 
+export const getOS = () => {
+    if (!navigator || !navigator.userAgent) return 'unknown';
+    const browser = Bowser.getParser(navigator.userAgent);
+    return browser.getOS().name.toLowerCase();
```
matejkriz: `name` is defined as possibly undefined (probably could never happen, but better safe than sorry)

szymonlesisz: addressed in 4270cee

also a lot of tests were added

src/ts/types/__tests__/index.ts
```diff
@@ -87,6 +87,13 @@ export const events = async () => {
             payload.features.unfinished_backup;
             payload.features.no_backup;
             payload.features.model;
+            // @ts-ignore: error does not exist
+            payload.error.toLowerCase();
```
matejkriz: maybe this would work in TS file?
```suggestions
payload.error?.toLowerCase();
```

matejkriz (approved):

src/ts/types/__tests__/index.ts
szymonlesisz: `@ts-expect-error` it is, thanks

PR #840, opened 2021-07-01
tests: add health check tests

szymonlesisz:
- checking `extendedDependencies` in package.json, to [prevent this](https://github.com/trezor/connect/commit/05bf0b19363390fbbb5ca8e7df9a1ad30da57583) 
- checking `!fixup` commits, like in trezor-suite

### DIFF
--- a/.github/workflows/fixup_check.yml
+++ b/.github/workflows/fixup_check.yml
@@ -1,0 +1,11 @@
+name: Fixup check
+
+on: [pull_request]
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


matejkriz (approved):
Great, that could help a lot! 🎉  

I would prefer the name of the script to be something like `check-extended-dependencies` instead of a bit confusing `test:health`, but I'm fine even with that... Feel free to merge.

szymonlesisz:
i'm going to extend `health` job in the future (checking package version, checking yarn lock, etc)

szymonlesisz:
btw the name was inspired by trezor-suite job

PR #841, opened 2021-07-02
feat(cardano): streamed transaction signing

gabrielKerekes:
This is an accompanying PR to https://github.com/trezor/trezor-firmware/pull/1683. For more context, please read the description in the FW PR.

Since the output format changed - Trezor no longer returns the whole serialized transaction - backwards compatibility with older FW is kept by parsing witnesses and auxiliary data supplement from the serialized transaction returned by older FW. For this we needed to add the `borc` package, hope that's OK. 

`auxiliaryData.blob` has been replaced with `auxiliaryData.hash` and removed completely so that the public type wouldn't have to confusingly contain both `blob` and `hash`. As a result arbitrary auxiliaryData can't be sent to older Trezor FW now - this shouldn't be a problem as arbitrary auxiliary data aren't widely used - https://github.com/trezor/connect/pull/794#issuecomment-829166073. Catalyst registration is still supported for both old and new FW though.

I plan on updating `cardanoSignTransaction.md` after the review process.


gabrielKerekes:
I've force-pushed updates to be compatible with changes in the FW PR - https://github.com/trezor/trezor-firmware/pull/1683#issuecomment-875339398.

gabrielKerekes:
I have again force pushed updates to be compatible with the related FW PR. This should thus be ready for review.

gabrielKerekes:
I've updated `cardanoSignTransaction.md` docs and I've added changelog entries into `CHANGELOG.md` just so that it won't be forgotten - I'm not sure if there's a better way to set the changelogs at this point. https://github.com/trezor/connect/pull/841/commits/6bbe14ded585688803c22af10a07843250a625e5

mroz22 (changes requested):
Hello, thanks for contribution! It looks quite ok in general but there are imho still few things/todos:

1. Could you please rebase on the lastest develop and update messages.json and respective types? 
2. This PR is about to introduce breaking changes from the trezor-connect consumers perspective - we are no longer returning serialized transaction here. With that in mind:
a] would you see it as a problem to stop supporting legacy firmwares for Cardano by increasing min firmware range limit? By doing this we could greatly reduce the amount of code changes here and most importantly future maintanance costs. If this is not a way to go for you, how long do you think legacy firmwares need to be supported? 
b] can't we possibly use the opposite approach and return signed tx in legacy format instead? If I understand recent changes correctly clients now need to assemble transactions on their own and only check its signature against the result returned from trezor. How complicated it would be to implement this into trezor-connect? Could you maybe provide some link to a referential implementation of this from another poject?

szymonlesisz:
Hi @gabrielKerekes 
I agree with @mroz22 that we should avoid this breaking change in the response of `cardanoSignTransaction` and we should return serialized transaction instead (legacy response)

there might be implementations of TrezorConnect where we can break something easily. like:
https://github.com/input-output-hk/daedalus/blob/db86f39ade58189375bc35c24bc6ad5c9dcbf998/source/main/ipc/getHardwareWalletChannel.js#L763


I don't know if you guys know about each other, but there is also [PR with Cardano implementation](https://github.com/trezor/trezor-suite/pull/4144) in `trezor-suite` which i guess is also relying on the legacy response from sign transaction.


What would be the easiest way (without using multiple cardano dependencies) to serialize new response from Trezor into legacy `serializedTx`?

gabrielKerekes:
We're not in favour of returning the whole serialized transaction from Connect. There’s multiple reasons for that:
1. Connect would have to receive as parameters additional fields which wouldn’t get passed into Trezor. E.g. native-scripts which are added in their raw form to the transaction witnesses (native scripts are not yet supported but are a part of multi-sig changes for which a PR has already been for FW), or auxiliary data of which Trezor receives only the hash. So data would get passed into connect which aren’t necessary for Trezor for the transaction signing itself and we would have to deal with them.
2. Alonzo era updates are coming to Cardano which will introduce new complex data into transactions (mainly outside of the tx body) and we want to keep these data away from HW wallets as they cannot verify/properly understand the data. If Connect would be serializing the transaction, it would also need to process this data even though it’s not required for the signing. These data might also change more frequently than the tx body and thus we would end up having to update Connect without any changes happening in the FW.
3. Introducing serialization into Connect means more code which might means greater potential for bugs.
4. Shouldn’t Connect be just a thin layer between the client and Trezor? Serializing the transaction is IMHO too much work for that.
5. Clients should verify that the transaction returned from Trezor is actually the same as the one the client wanted to sign i.e. the client should also calculate the transaction hash. Which means that the client needs to serialize the transaction anyways, which would result in unnecessary duplication. (I wasn’t able to find this in the Daedalus repo, but Adalite does it https://github.com/vacuumlabs/adalite/blob/develop/app/frontend/wallet/shelley/shelley-trezor-crypto-provider.ts#L415). However, an alternative to this could be parsing the serialized tx returned by Trezor and verifying its contents.
6. The three large wallets - Daedalus, Adalite and Yoroi integrate with both Ledger and Trezor. Ledger has always been returning just the witnesses of a transaction. Thus the migration to doing the same with Trezor shouldn’t be too much of a problem. (Yes, I’m aware this doesn’t apply to Trezor Suite and the PR would require significant work)
7. Daedalus as well as Yoroi developers have been informed of this breaking change at the beginning of July.

When it comes to legacy FW support, I would say we want to keep it for at least 6 months. Do you perhaps have any data regarding what FW versions are being used? Are users sceptical of updating to newer versions?

>I don't know if you guys know about each other

We don't.

gabrielKerekes:
Deadalus already has a [PR](https://github.com/input-output-hk/daedalus/pull/2629) ready for the changes.

gabrielKerekes:
We also just realised that if this PR would get merged too close to the FW release integrators might not have enough time to integrate properly. So far it seems major stakeholders should be ready for the release but it would be best to finalise this as soon as possible anyways. Otherwise we might need to take out the streaming changes from FW release in order not to break it.

cc @matejcik

gabrielKerekes:
I've rebased on top of develop and fixed the messages based on the current FW master as the common submodule doesn't contain the changes. https://github.com/trezor/connect/pull/841/commits/335ef4e5182a6a47d57111115e7b05a289d027d8

Hm, apparently this fails the CI now. How should the protobuf messages be updated then?

szymonlesisz:
if you have firmware repository in your filesystem, next to trezor-connect, run `./scripts/protobuf-build.sh local` this will use `../trezor-firmware/common/` instead of `trezor-common` submodule

gabrielKerekes:
> if you have firmware repository in your filesystem, next to trezor-connect, run ./scripts/protobuf-build.sh local this will use ../trezor-firmware/common/ instead of trezor-common submodule

I didn't notice the `local` option, but this is basically what I have done.

tsusanka:
Upon thorough discussion we agree to keep the backwards compatibility as you suggest. It would bring too much pain and effort to coordinate everything altogether. So we are fine with that.

However, please acknowledge that you will remove this legacy code in let's say 6 months and you will coordinate with the aforementioned wallets that they will migrate their code. Do you agree? Let's say end of February.

gabrielKerekes:
We will coordinate with the wallets to migrate their code. We're OK with removing the code in 6 months, but just to be sure, I'll confirm this with IOHK although I think it also shouldn't be a problem for them. Before I do that though, could you please answer my previous questions, just so we know?

> Do you perhaps have any data regarding what FW versions are being used? Are users sceptical of updating to newer versions?

tsusanka:
> Do you perhaps have any data regarding what FW versions are being used? Are users sceptical of updating to newer versions?

Vast majority is on latest. I'll share some numbers via email.

gabrielKerekes:
[It seems Daedalus supports Trezor FW >= 2.4.0](https://github.com/input-output-hk/daedalus/blob/dbed3018bdfb534f2e01bf3cdf226cefce8dca1e/source/renderer/app/config/hardwareWalletsConfig.js#L77). As most users are updating their FW I think it's safe to remove the legacy support in the 6 months. So we will remove the legacy code at the latest in ~6 months, end of February 2022.

There is however another issue. If we are not mistaken, releasing this version of Connect will result in downtime for integrators using the hosted version of Connect, unless (and probably even if) we coordinate carefully during the release or unless the major version of Connect is increased - I'm not sure if that's an option. Is there another way to circumvent this?

src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -130,21 +147,66 @@ export default class CardanoSignTransaction extends AbstractMethod {
...
+            networkId: payload.networkId,
+            witnessPaths: this._gatherWitnessPaths(
+                inputsWithPath,
+                certificatesWithPoolOwnersAndRelays,
+                withdrawals,
+            ),
         };
     }
 
+    _gatherWitnessPaths(
```
szymonlesisz: can we move it to one of the `./helpers/cardano...` files?

src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -155,38 +217,159 @@ export default class CardanoSignTransaction extends AbstractMethod {
...
+            for (const poolOwner of poolOwners) {
+                await typedCall('CardanoPoolOwner', 'CardanoTxItemAck', poolOwner);
+            }
+            // eslint-disable-next-line no-restricted-syntax
+            for (const poolRelay of poolRelays) {
+                await typedCall('CardanoPoolRelayParameters', 'CardanoTxItemAck', poolRelay);
+            }
+        }
+        // withdrawals
+        // eslint-disable-next-line no-restricted-syntax
```
szymonlesisz: lets allow `for(x in z)` and `for(x of z)` syntax globally.
please put this into .eslintrc "rules" and remove multiple eslint-disable lines from here 
```
"no-restricted-syntax": ["error", "LabeledStatement", "WithStatement"] // airbnb-base: allow ForInStatement and ForOfStatement
```

szymonlesisz (changes requested):
i've tested in on both 2-master and 2.4.0, works perfectly 
i have 2 little nitpicks

szymonlesisz:
> result in downtime for integrators using the hosted version of Connect

changing major version will not help us. If you have newest firmware but you are using old self-hosted connect then there is nothing we can do about it.

Luckily it will "only" end up with runtime error instead of coin loss, i guess we can live with that

src/js/core/methods/CardanoSignTransaction.js
gabrielKerekes: It doesn't really fit any of the other helper files so I moved it to a new one. https://github.com/trezor/connect/pull/841/commits/1bf70189bef41e0917bca2c4c2eb6370e46bcd48

src/js/core/methods/CardanoSignTransaction.js
gabrielKerekes: Done. https://github.com/trezor/connect/pull/841/commits/1bf70189bef41e0917bca2c4c2eb6370e46bcd48

gabrielKerekes:
I wasn't able to run the tests, but all the other checks pass and nothing really changed. For some reason `PYOPT=0 make build_unix_frozen` fails on the FW master branch via nix-shell and so I can't run the tests via trezor-user-env.

> Luckily it will "only" end up with runtime error instead of coin loss, i guess we can live with that

Ok, we'll have to think about the best way to do the switch.

refi93:
> changing major version will not help us. If you have newest firmware but you are using old self-hosted connect then there is nothing we can do about it.

The idea was that this PR (backwards compatible with current fw) would probably be released before the new firmware is out (indeed, it has to) - therefore, provided this gets launched as a new major version of connect, the integrating wallets would have time to switch to new connect major version witnout experiencing any runtime errors/downtime of Trezor Cardano transaction signing on their end, so in that sense, I think having this as a major version would indeed help. Or am I missing some other implications?

szymonlesisz:
>The idea was that this PR (backwards compatible with current fw) would probably be released before the new firmware is out (...)  integrating wallets would have time to switch to new connect major version

afaik FW will be released next wednesday, even if we publish those changes today i dont think 1 week is enough

> witnout experiencing any runtime errors/downtime of Trezor Cardano transaction signing on their end

If you update your FW to latest and then call `cardanoSingTransaction` on the __old__ implementation of TrezorConnect expecting legacy result in return, you will  end up with error anyway.
This is why i dont think changing major version will help.

Second thing is the maintenance of TrezorConnect.
Since we are hosting only `major versions` (connect.trezor.io/[major]) it means another branch/release to look up for us "only" because of `cardanoSignTransaction` method (which will not work anyway in conditions described above)
Unfortunately we are not able to force devs to update their dependency, therefore we still having implementations of version 4-7 all over the internet.
This is why are trying to save/spare "major versions" (hosting endpoints) for really big breaking changes. Otherwise in case of some global fixup (happens once in a while that we need to update something on old versions as well) we will be forced to maintain X more versions.

Leading/known Cardano implementations are aware about this change and prepared, so i guess we can release it as `8.2.0` (endpoint will be hosted on connect.trezor.io/__8__/, response from cardanoSingTransaction will be in __new__ format)

What do you think?

refi93:
@szymonlesisz so I gather a major version is not really an option given the overhead. In such case - could you tell us when exactly (day, hour) will you release the 8.2.0 version so we can coordinate the release of the changes on our end (and other web-based wallets as well) and minimize the period during which Trezor transactions won't work and notify users ahead of time?

Web-based wallets I know of (Adalite, Yoroi) are planning to work on the change this week (it's actually a small change given that Ledger, which these wallets support as well, already returns the signatures in a similar format), therefore the later you release it the better, as long as it is before the new trezor firmware release, of course. Something like September 7th would probably be ideal, supposing the trezor fw will be released on September 8th.

refi93:
Is there anything blocking the approval of this PR we can help with @szymonlesisz? If not, as asked above, could we get en ETA on the release of this change (ideally as precise as possible) to coordinate the release of the changes needed on our end?

gabrielKerekes:
I noticed I've forgotten to add the `.eslintrc` changes. I added them in https://github.com/trezor/connect/pull/841/commits/fb241abb074db155758347f0467373260eb7ce84

szymonlesisz:
@refi93 

Wednesday 8.09 at 16:00 CEST will be ok for you?

@gabrielKerekes 
fixup commit 335ef4e5182a6a47d57111115e7b05a289d027d8 looks wrong can we drop it? (removing "EthereumSignTxEIP1559" which is already in develop and removing "Lisk" which is [done here](https://github.com/trezor/connect/pull/879) both changes are not related to Cardano)

gabrielKerekes:
> fixup commit 335ef4e looks wrong can we drop it?

If the changes are already done elsewhere, then yes. I think I just wanted to make everything work back then.

refi93:
> Wednesday 8.09 at 16:00 CEST will be ok for you?

@szymonlesisz Yes that's fine, so that's the final time, right?

szymonlesisz:
@refi93 , yes it's final. we will release new trezor-connect together with new firmware

gabrielKerekes:
@szymonlesisz Daedalus people are having problems running production builds which are using a [pre-release](https://github.com/vacuumlabs/connect/releases/tag/8.1.30-rc.1-tx-streaming-extended) of this `cardano-streamed-transaction-signing` branch. It's probably caused by the addition of the `borc` package. Dev builds are working fine for them. 

Do you perhaps have any idea what could be causing the problem? Maybe I've built the package wrong? Or something else has to be done when adding `extendedDependencies`? To build the `tar.gz` in the pre-release I simply ran `yarn build:npm-extended` and then `yarn pack` in the `npm-extended` folder (the only modification I made in that pre-release in comparison with this PR is that I removed Lisk). [Here's the Deadalus PR](https://github.com/input-output-hk/daedalus/pull/2629).

I've tested the pre-release extended `tar.gz` with Cardano HW CLI and it seems to be working fine.
Screenshot 2021-09-03 at 10 06 10

gabrielKerekes:
Should I drop https://github.com/trezor/connect/commit/335ef4e5182a6a47d57111115e7b05a289d027d8 and rebase this on top of https://github.com/trezor/connect/pull/879?

szymonlesisz:
> Or something else has to be done when adding extendedDependencies? 
im not sure to be honest.
from what i saw in your pre-release package `borc` is listed in package.json so this should not be a problem, everything looks legit at the first sight.

i will try build and use this branch locally 

> Should I drop 335ef4e and rebase this on top of #879?

yes please, this way all tests (cardano + lisk) should work

szymonlesisz:
@gabrielKerekes, regarding that dependency error in Deadalus you were asking yesterday.

i just created empty simple node project, add npm-extended as dependency (from local fs) and successfully run script below without error.

EDIT: web build also doesnt throw any runtime error

```javascript

const TC = require('trezor-connect');

TC.default.manifest({
    email: 'a',
    appUrl: 'a',
});

TC.default.on(TC.UI_EVENT, evt => {
    console.log(evt);
    if (evt.type === TC.UI.REQUEST_PASSPHRASE) {
        TC.default.uiResponse({
            type: TC.UI.RECEIVE_PASSPHRASE,
            payload: {
                value: '',
                save: true,
                passphraseOnDevice: false,
            },
        });
    }
});

TC.default
    .cardanoSignTransaction({
        inputs: [
            {
                path: "m/44'/1815'/0'/0/1",
                prev_hash: '1af8fa0b754ff99253d983894e63a2b09cbb56c833ba18c3384210163f63dcfc',
                prev_index: 0,
            },
        ],
        outputs: [
            {
                address: 'Ae2tdPwUPEZCanmBz5g2GEwFqKTKpNJcGYPKfDxoNeKZ8bRHr8366kseiK2',
                amount: '3003112',
            },
        ],
        fee: '42',
        ttl: '10',
        protocolMagic: 764824073,
        networkId: 1,
        signingMode: 0,
    })
    .then(console.log);
    
```

gabrielKerekes:
@szymonlesisz Thanks for trying it out. We're investigating further. Currently we're testing out using the same version of borc as Daedalus uses.

szymonlesisz:
i can merge it today and publish connect with beta tag if you want (beta == extended) so you can try it out with "official npm pkg" instead of custom, locally created

gabrielKerekes:
If it wouldn't too much of a hassle to add potential fixes later today or tomorrow, then sure, let's merge and try 👍

szymonlesisz:
i've cherry-picked your commits into develop and released beta to npm registry

https://www.npmjs.com/package/trezor-connect/v/8.1.30-beta.3

gabrielKerekes:
The beta release also didn't work. @refi93 has tried replacing borc with cbor [here](https://github.com/vacuumlabs/connect/commit/6783387c4364e2193c7117cbe33f6b5ee1589b76). Daedalus have confirmed the replaced version works. I'll test it once more today/early tomorrow and I will add it to this PR.

szymonlesisz:
i guess this fix will require some rebase over  develop branch (this PR is already outdated)

@refi93 pls move "cbor" to dev dependencies like it was with borc (i know it's crazy :)

gabrielKerekes:
I've created a separate PR for it: https://github.com/trezor/connect/pull/884 Cbor is moved to dev dependencies. Everything appears to be working as before. Connect integration tests pass for current and old FW. I also tested with Cardano HW CLI. It would be awesome if you could create another beta release before noon tomorrow so Daedalus could do their QA. 🙏

szymonlesisz:
@refi93 i've merged `cbor` PR but now web build is failing on:
```
ERROR in ./node_modules/cbor/lib/utils.js
Module not found: Error: Can't resolve 'node-inspect-extracted' in '/Workspace/trezor-connect/node_modules/cbor/lib'
```

all happens because of this:

Screenshot from 2021-09-08 11-07-33


also i've noticed that added `cbor` package is already outdated, there is no such code in cbor@8.0.0 
https://github.com/hildjj/node-cbor/blob/main/packages/cbor/lib/utils.js


is there any reason why we are not using 8.0.0?

they recommend to use this package for web builds:
https://github.com/hildjj/node-cbor/tree/main/packages/cbor-web

szymonlesisz:
> is there any reason why we are not using 8.0.0?

i will answer to it myself :)
```
error cbor@8.0.0: The engine "node" is incompatible with this module. Expected version ">=12.19". Got "12.18.4"
```

gabrielKerekes:
I'm not sure about the exact difference between `cbor-web` and `cbor`. Is there a way we can use different package for different builds? Or should we just use `cbor-web`? Sadly I don't have enough experience with JS builds.

gabrielKerekes:
It's mentioned on the [cbor-web npm page](https://www.npmjs.com/package/cbor-web) that:
> You can require('cbor-web') from node.js, but I wouldn't recommend that unless you're trying to use the exact same paths for backend and frontend codebases, for example

The part after "unless" is exactly our case, is it not? What do you think @szymonlesisz? I tried building the extended package and using it in Cardano HW CLI and it's working just fine.

gabrielKerekes:
I've created a PR using the `cbor-web` package -> https://github.com/trezor/connect/pull/886.

szymonlesisz:
im getting:
```
ReferenceError: TextDecoder is not defined
```
in nodejs tests (cbor@8.0.0 and/or cbor-web@8.0.1)

gabrielKerekes:
I've used `cbor-web@^7.0.6`.

szymonlesisz:
yeah, but i'm not a big fan of fixing on outdated dependency version , im still trying to figure this out, how to use 8.0.0

refi93:
> yeah, but i'm not a big fan of fixing on outdated dependency version , im still trying to figure this out, how to use 8.0.0

@szymonlesisz Some projects depending on connect may still use node v12 (still maintained) but a version lower than 12.19 which is apparently required by cbor 8.0.0 (https://github.com/hildjj/node-cbor/blob/b670b3ab5d39a7ca4a6ea168e12e877721871881/package.json#L51) - so IMHO that can be a valid reason to stick to version 7

szymonlesisz:
i guess you are right, another thing that this is only temporary solution (legacy responses) which will be removed in the future.

im merging cbor-web PR then

gabrielKerekes:
If everything's green could you then also create another extended beta release please?

szymonlesisz:
still having issues on web build. now im getting runtime error while calling `cardanoSignTx` with older FW:

`cbor_web__WEBPACK_IMPORTED_MODULE_1___default.a is undefined`
​​

EDIT: it needs to be `import * as cbor from 'cbor-web';`

ive fixed it here https://github.com/trezor/connect/commit/cf7b4ed10f2b3c0100806e1ba03b34d34878e605

npm build is on the way

szymonlesisz:
beta.4 published

https://www.npmjs.com/package/trezor-connect/v/8.1.30-beta.4

szymonlesisz:
i've published another beta version
https://www.npmjs.com/package/trezor-connect/v/8.1.30-beta.5

firmware releases are now stored in new @trezor/connect-common dependency.

it should not break anything in your implementation, but can you try it just to be sure?

refi93:
@szymonlesisz works fine on web, but actually 8.1.30-beta.5 on node results in the following error:

```
Error: ENOENT: no such file or directory, open '/home/rafael/workplace/cardano/cardano-hw-cli/node_modules/node_modules/@trezor/connect-common/files/bridge/releases.json'
```

8.1.30-beta.4 works fine

refi93:
It seems that `/node_modules` is included in the path twice

refi93:
> > Wednesday 8.09 at 16:00 CEST will be ok for you?
> 
> @szymonlesisz Yes that's fine, so that's the final time, right?

@szymonlesisz is it released already? I'm checking with `curl -s https://connect.trezor.io/8/trezor-connect.js | grep VERSION` and it seems to still host 8.1.29 ATM

szymonlesisz:
no, sorry we just fixed this error introduced by beta.5 (double node_modules)
https://www.npmjs.com/package/trezor-connect/v/8.1.30-beta.6

now should work

szymonlesisz:
could you pls confirm that it's working for you? if it's working then we are ready to ~~release now~~

EDIT: shit, sorry guys i was just notified that there is some problem with the release of trezor-suite (not connect) therefore new  firmware will not be released today, therefore connect should not be released (because now it's offering you new, not released FW)

We will let you know asap (tomorrow) about the status, we are promising that we will **not** release trezor-connect without your ACK

refi93:
regarding the issue with `borc` on Daedalus, turns out their production build script had to explicitly include sub-dependencies and ieee754 nor cbor-web were among them, so it wasn't an issue neither with borc nor connect: https://github.com/input-output-hk/daedalus/pull/2629/commits/0675c2e0f256d7ee33b2132f0fcd715bbc2a1b24 - the current version of the package with cbor-web works fine as they fixed it.

So no further action/investigation is needed, sorry for the hassle

refi93:
@szymonlesisz do you have any ETA on the connect release? We would like to inform also other web wallets (Yoroi, ccwallet.io) so they have time to make the switch on their side

tsusanka:
We will release Connect at around 15:00 Sep 15th and ping you when done.

zuzunker:
> We will release Connect at around 15:00 Sep 15th and ping you when done.

Hi! Is that 15:00 UTC?

Scitz0:
Hi, I'm working part-time on ccwallet.io and I am tasked to make sure our Trezor implementation is up-to-date for the new FW. I haven't really kept myself in the loop of the changes but browsed through the conversation here and I think I got the gist of it. I have left it to the last minute but shouldn't be too hard to get it in place in time I hope 😊

Some questions though to get up to speed quicker. 

* We should implement changes for, and have 8.1.30 deployed before, at, or after 15:00 Sep 15th?
* Beta 7 is the one to go for or will there be a final version before the above date?
* What is the FW version of the upcoming Trezor release?
* Is it recommended to enforce FW upgrade as we have CIP20 implemented to add messages/comments/memos as transaction metadata by using the metadatum label 674 as arbitrary auxiliary data? Old FW is not compatible with this if I understand the note from the first message correctly?
> auxiliaryData.blob has been replaced with auxiliaryData.hash and removed completely so that the public type wouldn't have to confusingly contain both blob and hash. As a result arbitrary auxiliaryData can't be sent to older Trezor FW now - this shouldn't be a problem as arbitrary auxiliary data aren't widely used

tsusanka:
@vsubhuman apologies, I meant 15:00 CET, that is 13:00 UTC.

@gabrielKerekes do you think you could answer @Scitz0 questions?

gabrielKerekes:
@tsusanka please correct me if I am wrong in the release related answers.

> - We should implement changes for, and have 8.1.30 deployed before, at, or after 15:00 Sep 15th?
> - Beta 7 is the one to go for or will there be a final version before the above date?

You can implement changes against Beta 7 and switch to the official, non-beta, package after it's released at 15:00.

> What is the FW version of the upcoming Trezor release?

2.4.2

> Is it recommended to enforce FW upgrade as we have CIP20 implemented to add messages/comments/memos as transaction metadata by using the metadatum label 674 as arbitrary auxiliary data? Old FW is not compatible with this if I understand the note from the first message correctly?

Yes, you should enforce FW upgrades in order to support auxiliary data.

Scitz0:
Thanks 👍  
I think I have the required changes in place now to be deployed tomorrow.

tsusanka:
Hello everyone, we will be releasing Connect in number of minutes. The pipeline is running. We will ping here when it is out.

Important note: we are releasing Connect but we are not yet releasing the new firmwares (due to some unfortunate circumstances). That however does not prevent anything, you can do your Connect migration and once the new firmwares are out (should be next week at latest) it should work out of box.

tsusanka:
It is out!

tsusanka:
We have deployed both to NPM and to connect.trezor.io. I am closing here as these changes are already in `develop` and deployed. Please let us know either here or open a new issue in case of some troubles!

Scitz0:
Wasn't this supposed to be synced with Trezor FW 2.4.2, or did I get that wrong?
Trezor Suite says that the device is up-to-date (2.4.1), no newer FW available.

tsusanka:
Sorry for confusing you. At first yes, but as we ran into some troubles with FW release, we have decided to at least release Connect, so you all are not dependent on our release date. As the Connect changes are backwards compatible with 2.4.1 you can do the "migration" already. And once 2.4.2 is out it should work out of the box.

Scitz0:
> Sorry for confusing you. At first yes, but as we ran into some troubles with FW release, we have decided to at least release Connect, so you all are not dependent on our release date. As the Connect changes are backwards compatible with 2.4.1 you can do the "migration" already. And once 2.4.2 is out it should work out of the box.

Ok, I updated our code to utilize the new format as seen in:
https://github.com/vacuumlabs/connect/blob/e7d2ca29a8069362341dc6058b4482be54853b81/docs/methods/cardanoSignTransaction.md
When testing it out I got:
`Feature AuxiliaryDataHash not supported by device firmware`
So I guess this will be broken until 2.4.2 is released then. Other than that tx was successfully sent using updated code if I omit metadata.

tsusanka:
New firmwares are released :tada:.

Scitz0:
> New firmwares are released 🎉.

Tested and confirmed, general metadata working again 💪

PR #844, opened 2021-07-09
Chore/update protobuf messages

szymonlesisz:
new messages taken from trezor-firmware repo, commit [66bf309fbf98d3c3d8209285bf40e7e4a33c47cc](https://github.com/trezor/trezor-firmware/commit/66bf309fbf98d3c3d8209285bf40e7e4a33c47cc)
main changes: added "pages" and "page_number" fields in ButtonRequests

- expose whole ButtonRequest object in UI.REQUEST_BUTTON event (previously only "code" was exposed)
- `ButtonRequestMessage` and `DeviceMessage` (events) are now strongly typed by enum from protobuf, (previously it was `string`)

### DIFF
--- a/src/data/messages/messages.json
+++ b/src/data/messages/messages.json
@@ -1726,7 +1726,8 @@
                 }
             ],
             "options": {
-                "(wire_type)": 22
+                "(wire_type)": 22,
+                "(include_in_bitcoin_only)": true
             },
             "oneofs": {}
         },
@@ -1761,7 +1762,8 @@
                 }
             ],
             "options": {
-                "(wire_type)": 22
+                "(wire_type)": 22,
+                "(include_in_bitcoin_only)": true
             },
             "oneofs": {}
         },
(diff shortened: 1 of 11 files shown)


tsusanka:
@tomasklim we have decided to merge this so we can put it into Suite so we can put it on staging so we can test in QA. Please review "ex-post".

src/data/messages/messages.json
```diff
@@ -8950,14 +8993,21 @@
...
                     "name": "wire_type",
-                    "id": 50002
+                    "id": 52002
+                },
+                {
+                    "rule": "optional",
+                    "options": {},
+                    "type": "bool",
+                    "name": "include_in_bitcoin_only",
+                    "id": 60000
```
tomasklim, line 9010: Do you know what was the reason that ids were changed in fw? And also that until the last update it was a sequence from 5000 to 5008 and now it is jumping by hundreds?

tomasklim (commented):
LGTM.

PR #846, opened 2021-07-14
Fix/dispose method

### LINKED ISSUE #515: Node support - clean up
I could not find a way to clean up after I am done with trezor, and there's leaking connection to the bridge server. Even if you clean up everything in trezor-connect, there will still be fetch request to `/listen` and currently there's no way to pass `signal` from `Abortcontroller`, so the process will get stuck unless device connect/disconnect event happens. I could not think of good way to do this with small changes, other than "hacking" `fetch`, collect abortion signals for requests and do clean up afterwords or some other ugly solutions.

`trezor-link` v2 bridge needs to track `listen` request and have its abort controller in order to stop when `stop` is called on it. (node-fetch also supports AbortController`)

Will be nice to have clean in up tools on node.js, without it process wont exit or clean up without hard exit. (`process.exit`)

szymonlesisz:
- fixed `TrezorConnect.dispose` and allow tests to ends gracefully. (`--detectOpenHandles --forceExit` are not longer required) fix: #515

- rename `onBeforeUnload` methods to `dispose`

### DIFF
--- a/src/js/env/node/index.js
+++ b/src/js/env/node/index.js
@@ -38,7 +38,7 @@ export const manifest = (m: $T.Manifest) => {
 export const dispose = () => {
     eventEmitter.removeAllListeners();
     if (_core) {
-        _core.onBeforeUnload();
+        _core.dispose();
     }
     // TODO: fix top level values during refactor
     // $FlowIssue
--- a/src/js/env/node/index.js
+++ b/src/js/env/node/index.js
@@ -36,10 +36,15 @@ export const manifest = (m: $T.Manifest) => {
 };
 
 export const dispose = () => {
-    // iframe.dispose();
-    // if (_popupManager) {
-    //     _popupManager.close();
-    // }
+    eventEmitter.removeAllListeners();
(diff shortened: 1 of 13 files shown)


src/js/env/node/index.js
```diff
@@ -36,10 +36,15 @@ export const manifest = (m: $T.Manifest) => {
...
 export const dispose = () => {
-    // iframe.dispose();
-    // if (_popupManager) {
-    //     _popupManager.close();
-    // }
+    eventEmitter.removeAllListeners();
+    if (_core) {
+        _core.dispose();
+    }
+    // TODO: fix top level values during refactor
```
matejkriz, line 43: What this TODO means, please? What do you mean by those "top level values"?

szymonlesisz: i meant variables :)

_core is a top level variable in this module and this is why // $FlowIssue was required

it needs to be refactored properly, but it is out of scope for this PR

matejkriz (approved):

PR #847, opened 2021-07-14
feat(ci): automate production and beta releases to npm registry

vdovhanych:
This pr will allow us to publish releases to the npm registry with a simple button press. We will have 2 jobs one for production running only on v8 branch and beta one for all other branches with an option to upload beta tag packages.

### DIFF
--- a/.github/workflows/version_check.yml
+++ b/.github/workflows/version_check.yml
@@ -1,0 +1,27 @@
+name: "connect version check"
+on:
+  pull_request:
+    branches:
+      - v8
+jobs:
+    version_commit_check:
+      name: Check connect version commit message
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - run: |
+            LAST_COMMIT=$(git log -1 --pretty=%B)
+            VERSION=$(jq -r .version package.json)
+            [[ "$LAST_COMMIT" == "version $VERSION" ]]; if [ $? -eq 0 ]; then echo "Last commit message contains version";   exit 0; else echo "Last commit message DOES NOT contain version" >&2;   exit 1; fi
+
+    version_bump_check:
+      name: Check connect version bump
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - uses: actions/setup-node@v2
+        - run: |
+            yarn install
+            REMOTE_VERSION=$(npm show trezor-connect version)
+            LOCAL_VERSION=$(jq -r .version package.json)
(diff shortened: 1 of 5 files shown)


szymonlesisz (approved):

vdovhanych:
I have done some checks that @szymonlesisz  was asking for but I'm not really sure it's what he wanted. Check it if it's something you think we can work with and let me know. Yes, there is one small script in js that will check the local and remote versions, I think it's really not that great from me so I would appreciate any suggestions :)

scripts/ci-compare-version.js
```diff
@@ -0,0 +1,6 @@
+var compareVersions = require('compare-versions');
```
tomasklim:
```suggestion
const compareVersions = require('compare-versions');
```

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    version_bump_check:
+      name: Check connect version bump
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - uses: actions/setup-node@v2
+        - run: |
+            npm install
+            npm i compare-versions
```
tomasklim: add to `package.json`

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,28 @@
...
+            [[ "$LAST_COMMIT" == "version $VERSION" ]]
+
+    version_bump_check:
+      name: Check connect version bump
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - uses: actions/setup-node@v2
+        - run: |
+            npm install
```
tomasklim: we use `yarn`

tomasklim (changes requested):

tomasklim (approved):

tsusanka:
@vdovhanych squash and force push? And let's merge

tsusanka:
See here why the CI fails. https://github.com/trezor/connect/pull/848#issuecomment-883409848 It is not related to this PR.

tsusanka (approved):
LGTM, though @vdovhanych haven't we agreed we will prefer Gitlab over Github Actions? :) But let's leave it like this for now, it is OK

vdovhanych:
Yes, @tsusanka I wanted to do it in Gitlab CI but it was not possible to do what we wanted. That is to run these checks on a pull requests to a v8 branch and as we do not push into the GitLab there is no other way. If we had these checks on Gitlab they would either run all the time which would mean that the pipeline would fail most of the time or we would have them run on push to a v8 branch but that doesn't provide us with preventing incorrect versioning in the v8 branch.

PR #848, opened 2021-07-15
fix: TrezorConnect.firmwareUpdate method

marekrjpolak:
- uploaded binary modified in the same way as when downloaded by version
- Rollout version 1.2.0 needed

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
         "@babel/preset-env": "^7.12.10",
         "@babel/preset-flow": "^7.12.1",
         "@trezor/blockchain-link": "^1.0.17",
-        "@trezor/rollout": "1.1.0",
+        "@trezor/rollout": "^1.2.0",
         "@trezor/utxo-lib": "^0.1.2",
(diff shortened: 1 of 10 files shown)


marekrjpolak:
Looks good, everything I need is there.

mroz22 (approved):
looks good

tomasklim (approved):

tsusanka:
I think this will conflict with https://github.com/trezor/trezor-suite/pull/4038/files. Which one should we merge first?

tsusanka:
Gitlab is failing because someone sent money to [this address](https://etherscan.io/address/0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8) and we check its balance. There is already [pending tx](https://etherscan.io/tx/0x50ca4d1086ec7cc6b17ab57658993da84e2ae22770ee417bea22303c518fe45c) that will solve this so I am gonna ignore it for now but it is not a good idea to depend on such externality and someone could troll us very easily this way. I believe we should mock this.

PR #853, opened 2021-07-21
feat(ci): add job for deploying connect to connect.trezor.io

vdovhanych:
This pr will allow us to connect directly to connect.trezor.io right from your GitLab ci with a manual action. This job will only run on the v8 branch. It is still a work in progress as we need to add some prerequisites to the GitLab CI environment, but otherwise, it should be good to go.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -88,8 +88,6 @@ publish release to npm:
 publish beta release to npm:
   image: node:14
   stage: deploy
-  except:
-    - v8
   when: manual
   before_script:
     - echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}">.npmrc
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,8 +26,8 @@ build:
     paths:
       - build
 
-# Deploy
-deploy review:
+# Deploy to testing environment
+deploy test:
   stage: deploy
   variables:
     GIT_STRATEGY: none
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -26,7 +26,7 @@ build:
     paths:
       - build
 
-# Deploy
+# Deploy to testing environment
 deploy review:
```
tsusanka: Let's rename this to `deploy test`?

.gitlab-ci.yml
```diff
@@ -49,6 +49,23 @@ deploy review:
   tags:
     - deploy
 
+# Deploy release to connect.trezor.io
+deploy to connect.trezor.io:
```
tsusanka: And this to `deploy production`?

.gitlab-ci.yml
```diff
@@ -49,6 +49,23 @@ deploy review:
...
+  stage: deploy
+  only:
+    refs:
+      - v8
+  when: manual
+  variables:
+    GIT_STRATEGY: none
+  before_script:
+    - make clean
+    - make build
```
tsusanka: No reason to have this in `before_script`, we can have it in `script` directly, nope? Yes, bikesheding 101.

tsusanka (approved):

tsusanka:
I am still a bit pessimistic the `make build` command will work out of box on the CI but we'll see :).

vdovhanych:
> I am still a bit pessimistic the `make build` command will work out of the box on the CI, but we'll see :).

Well, you're right. It won't work out of the box. We need still some perquisites for this to work. 
Base docker image (we can use the one we have on the suite, I guess) 
AWS CLI installed in that base docker
Adding deploy key and authenticate the AWS CLI with that.

vdovhanych:
We should be ready to merge and try this. All the necessary things are set up for the deployment.

vdovhanych:
I also added a small fix for publishing the beta release to npm. If needed, I can do the fix in a separate branch.

PR #867, opened 2021-08-04
Chore/remove makefile

szymonlesisz:
- move all `Makefile` scripts to bash and remove `Makefile` completely
- improved docker build (multiple targets)


vdovhanych (approved):

szymonlesisz:
i've force-pushed because of missing `[ $# -ge 1 ]` in bash scripts (effect of `set -u`)

tsusanka (approved):

PR #868, opened 2021-08-06
feat(ci): improve commit message check on pr to v8

vdovhanych:
Hi, this pull request introduces better commit message checking for pull requests to the `v8` branch. Now the check will let you merge if it contains one of 3 types of commits 
~~1. version x.x.x. (same as `package.json` version)~~
~~2. cherry-picked commits from develop with `git cherry-pick <develop hash> -x `(for message cherry-picked from ...)~~
~~3. with suffix `[RELEASE ONLY]`~~
The script will now only check if the last commit is indeed present in the `develop` branch if not it fails.

We can modify it more but I think it's what you wanted to add @szymonlesisz

### DIFF
--- a/scripts/ci-check_release_commit_messages.sh
+++ b/scripts/ci-check_release_commit_messages.sh
@@ -7,35 +7,17 @@ git fetch origin develop
 # list all commits between HEAD and develop
 for commit in $(git rev-list origin/develop..)
 do
-    message=$(git log -n1 --format=%B $commit)
+    COMMIT_ID=$(git log --pretty=format:'%H' -n 1 $commit)
     echo "Checking $commit"
-    version=$(jq -r .version package.json)
-    echo "Checking connect version"
 
-    # The commit message must contain either
+    # 1. Checking if the commit is in develop"
 
-    # 1. Version substring
-    if [[ $message == "version $version" ]]; then
-      continue
-    fi
-
-    # 2. "cherry-picked from [some commit in develop]"
-    if [[ $message =~ "(cherry picked from commit" ]]; then
-      # remove last ")" and extract commit hash
-      develop_commit=$(echo ${message:0:-1} | tr ' ' '\n' | tail -1)
-      # check if develop really contains this commit hash
-      if [[ $(git branch -a --contains $develop_commit | grep --only-matching "remotes/origin/develop") == "remotes/origin/develop" ]]; then
-        continue
-      fi
-    fi
-
-    # 3. [RELEASE ONLY] substring
-    if [[ $message =~ "[RELEASE ONLY]" ]]; then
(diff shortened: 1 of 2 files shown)


tsusanka:
As we always move the v8 branch forward for every release (not like with firmware/Suite releases where the branch stays). I am not sure option 2 and 3 makes sense as `v8` should not diverge from `develop` I guess (it only can be behind).

But I am leaving this up to @szymonlesisz.

vdovhanych:
Update: I changed the script only to check if the commit is indeed in develop. Other options aren't useful for this use case.

scripts/ci-check_release_commit_messages.sh
```diff
@@ -7,35 +7,17 @@ git fetch origin develop
 # list all commits between HEAD and develop
 for commit in $(git rev-list origin/develop..)
 do
```
tsusanka: Is the looping really needed? Because I thought the idea of `git merge-base` was to avoid this.

vdovhanych: Yep, you're right it's pointless now. Removed it.

scripts/ci-check_release_commit_messages.sh
```diff
@@ -0,0 +1,24 @@
...
+
+git fetch origin develop
+
+# list all commits between HEAD and develop
+commit=$(git rev-list origin/develop..)
+
+commit_id=$(git log --pretty=format:'%H' -n 1 $commit)
+echo "Checking $commit"
+
+# 1. Checking if the commit is in develop"
```
tsusanka, line 13: Let's remove the `1. `.

scripts/ci-check_release_commit_messages.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [[ $(git merge-base --is-ancestor $commit_id HEAD | grep --only-matching "remotes/origin/develop") == "remotes/origin/develop" ]]; then
+  continue
```
tsusanka, line 16: You are not in a loop, `continue` does not make sense here?

```
if ..; then
  echo ..
  exit 0
exit 1
```

And you can drop the `fail` variable altogether.

scripts/ci-check_release_commit_messages.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+fail=0
+
+git fetch origin develop
+
+# list all commits between HEAD and develop
+commit=$(git rev-list origin/develop..)
```
tsusanka, line 8: I do not think this is doing what the comment says.

tsusanka (commented):
Please read the script again. There are clearly leftovers from the previous design and I am not sure it even works now. @vdovhanych please slow down and be more thorough it is not a race :).

vdovhanych:
We decided that this is not really necessary and instead will check only if the version in package.json is not beta. Will create new pr.

PR #875, opened 2021-08-17
Fix/unavailable capabilites

szymonlesisz:
Motivation:
`config.json` fields and `device.unavailableCapabilities` were inconsistent

Problems:
- invalid detection - 'no-capability' (from device.features) was set before 'no-support' (from coins.json)
- there was no possibility to disable support (using "0" value) from config.json
- `device.unavailableCapabilities` returns literal string (reason) OR array of strings (not anymore)
- `config.json` supportedFirmware rules confusing field names (excludedMethods mixed with defined capability/feature)
- there was no tests for `getFirmwareRange` utility

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,8 @@
 webpack
 src
 build
+__tests__
+__fixtures__
 _old
 .babelrc
 .flowconfig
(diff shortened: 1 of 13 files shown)


marekrjpolak (approved):
Probably OK

tomasklim (approved):

PR #879, opened 2021-08-25
Chore/remove Lisk

szymonlesisz:
removing Lisk integration in reaction to https://github.com/trezor/trezor-firmware/pull/1765

we are not aware about any active trezor-connect integration with Lisk


marekrjpolak (approved):
LGTM

PR #883, opened 2021-09-07
feat: copy bridge and firmware releases from connect-common

matejkriz:


tsusanka:
This means we do not have to modify the releases.json here right? And we'll just change it in Suite in connect-common's `releases.json`.

matejkriz:
> This means we do not have to modify the releases.json here right? And we'll just change it in Suite in connect-common's `releases.json`.

exactly, this deletes the releases.json files from here to have the single source of truth in Suite repo

package.json
```diff
@@ -62,6 +62,7 @@
         "@babel/preset-env": "^7.12.10",
         "@babel/preset-flow": "^7.12.1",
         "@trezor/blockchain-link": "^1.0.17",
+        "@trezor/connect-common": "^0.0.2",
```
szymonlesisz, line 65: you need to add it to "extendedDependency" as well (to expose this dependency in npm-extended package)

src/js/env/node/networkUtils.js
```diff
@@ -8,13 +8,34 @@ if (global && typeof global.fetch !== 'function') {
...
+    __dirname,
+    '../../../../node_modules/@trezor/connect-common/files/',
+);
+
+const getAbsolutePath = (fileUrl: string) => {
+    switch (fileUrl) {
+        case './data/bridge/releases.json':
+            return path.join(connectCommonFilesPath, 'bridge/releases.json');
+        case './data/bridge/latest.txt':
+            return path.join(connectCommonFilesPath, 'bridge/latest.txt');
```
szymonlesisz: im wondering if this file is even used (there is no such case in react-native)

szymonlesisz (changes requested):

src/js/env/node/networkUtils.js
matejkriz: 👍 deleted

package.json
matejkriz: 👍 added

szymonlesisz (approved):

PR #885, opened 2021-09-08
Fix/cbor version

szymonlesisz:
fix for web build


szymonlesisz:
cbor-web will be used instead https://github.com/trezor/connect/pull/886

PR #887, opened 2021-09-08
feat(stellar): strong typing in stellar protobuf

matejcik:
companion PR for https://github.com/trezor/trezor-firmware/pull/1755

to recap:
* `StellarAssetType` is renamed to `StellarAsset`, and the name `StellarAssetType` is reused for an enum.
  * this causes no issue in Connect whatsoever, far as I was able to determine
* many fields are now required. These seem to mostly match what Connect already assumed, with some minor tweaks.
* some fields are now enums
* `timebounds` are now required, and it's an error to not specify them
  * this should be OK from Stellar point of view, as omitting timebounds is _strongly discouraged_. Stellar Account Viewer always specifies timebounds.

I was unable to get the typechecker to pass, so I could use some hints. At this point `flow` doesn't say anything more specific than "exact not matching inexact blabla"

### DIFF
--- a/src/data/messages/messages.json
+++ b/src/data/messages/messages.json
@@ -9370,12 +9370,12 @@
             "oneofs": {}
         },
         {
-            "name": "StellarAssetType",
+            "name": "StellarAsset",
             "fields": [
                 {
                     "rule": "required",
                     "options": {},
-                    "type": "uint32",
+                    "type": "StellarAssetType",
                     "name": "type",
                     "id": 1
                 },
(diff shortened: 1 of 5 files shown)


matejcik:
Fixed the type problems now.

All test vectors are currently failing because of the new requirements that timebounds are set.
In https://github.com/trezor/trezor-firmware/pull/1755 I exported test vectors from firmware into a common JSON, but they're in protobuf format, not Connect's input format.

matejcik:
I tried to use `stellar_sdk` Python library to statically generate testcases for Connect, but it didn't really help that much, the naming and other conventions are still quite different.

It might be possible to use javascript stellar SDK to parse the provided XDR in a format more closely matching what Connect expects.

matejcik:
https://github.com/trezor/trezor-firmware/pull/1755 is now in master and the test fixtures are exported to trezor-common.

Note that **this PR is not required for Stellar to work.**
Existing integrations should work fine with new firmwares, as long as timebounds are provided.
(I can modify Stellar testcases in Connect to provide timebounds, and the tests pass with no further changes to Connect code)

szymonlesisz:
closing via https://github.com/trezor/connect/pull/913

PR #888, opened 2021-09-08
feat(ci): add a job for making a rollback copy of connect

vdovhanych:
Add job for making a copy of production connect to rollback in case we need to revert production version to the previous one and a small fix to the release job of connect to aws.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -53,6 +53,22 @@ deploy test:
   tags:
     - deploy
 
+# Create rollback copy of connect.trezo.io
+rollback production copy:
+  stage: deploy
+  only:
+    refs:
+      - v8
+  when: manual
+  needs: ["build"]
+  variables:
+    GIT_STRATEGY: none
+  before_script: []
+  script:
+    - aws s3 sync s3://connect.trezor.io s3://rollback-connect.trezor.io
+  tags:
+    - deploy
+
 # Deploy release to connect.trezor.io
 deploy production:
   stage: deploy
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -53,6 +53,22 @@ deploy test:
   tags:
     - deploy
 
+# Create rollback copy of connect.trezo.io
+rollback copy:
```
tsusanka: Maybe `rollback production copy` is better?

tsusanka (approved):
I like it, I am wondering just one thing: everyone logged in will be able to trigger this action right? Because if someone would hit it by mistake we would lose the backup. On the other hand is that a realistic scenario? Also once connect is in Suite monorepo we could bind it to `codesign` I guess.

vdovhanych:
> I like it, I am wondering just one thing: everyone logged in will be able to trigger this action right? Because if someone would hit it by mistake we would lose the backup. On the other hand is that a realistic scenario? Also once connect is in Suite monorepo we could bind it to `codesign` I guess.

Yes i that is correct but it will only work on v8 branch and it could be hit by mistake only before release so i think it's not that big deal. As for the monorepo i agree that we could tie this to codesign so only select people will have access to this.

.gitlab-ci.yml
vdovhanych: yes that looks better, will change to it.

tsusanka: Feel free to force-push then.

tsusanka:
Ok very well then!

.gitlab-ci.yml
vdovhanych: done

tsusanka (approved):

PR #915, opened 2021-10-08
docs: example server-side verification for requestLogin in Golang

kaloyan-raev:
I implemented server-side signature verification for `requestLogin` in Golang using the [github.com/btcsuite/btcd/btcec](https://pkg.go.dev/github.com/btcsuite/btcd/btcec) package. This example might be helpful for other Golang developers.


tsusanka (approved):
Sorry for the delay, thanks for the contribution!

PR #920, opened 2021-10-15
feat(cardano): add support for multisig

davidmisiak:
This PR adds support for Cardano native scripts and token minting. Relevant firmware PR [here](https://github.com/trezor/trezor-firmware/pull/1772).

There should be no breaking changes. New features:
- `cardanoGetNativeScriptHash` call ([docs](../blob/276ad461774e78ad8d46d6c9138c2e2cae682b4b/docs/methods/cardanoGetNativeScriptHash.md))
- multi-sig transactions ([docs](https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/cardano/README.md#multi-sig-transaction))
  - [Script credentials](../blob/276ad461774e78ad8d46d6c9138c2e2cae682b4b/src/js/types/networks/cardano.js#L37-L45) must be used in certificates and withdrawals when signing a multi-sig transaction (using the `MULTISIG_TRANSACTION` signing mode). Ordinary (1852') witness requests are not allowed. All the multisig witness requests (1854') should be provided in the [`additionalWitnessRequests` field](../blob/276ad461774e78ad8d46d6c9138c2e2cae682b4b/src/js/types/networks/cardano.js#L188) in the `cardanoSignTransaction` call and they are all shown to the user. Multisig transaction cannot contain a pool registration certificate.
- token minting and burning ([docs](https://github.com/trezor/trezor-firmware/blob/master/core/src/apps/cardano/README.md#token-mintingburning))
  - There is a new optional [`mint` field](../blob/276ad461774e78ad8d46d6c9138c2e2cae682b4b/src/js/types/networks/cardano.js#L187) in the `cardanoSignTransaction` call. Keys derived from a minting path (1855') are used to create token minting/burning policies. Corresponding witness requests should be provided in the [`additionalWitnessRequests` field](../blob/276ad461774e78ad8d46d6c9138c2e2cae682b4b/src/js/types/networks/cardano.js#L188) in the `cardanoSignTransaction` call and they are all shown to the user.

### DIFF
--- a/src/js/types/networks/cardano.js
+++ b/src/js/types/networks/cardano.js
@@ -92,13 +92,15 @@ export type CardanoInput = {
 
 export type CardanoToken = {
     assetNameBytes: string,
-    amount: string,
+    amount?: string,
+    mintAmount?: string,
 };
 
 export type CardanoAssetGroup = {
     policyId: string,
     tokenAmounts: CardanoToken[],
 };
+
 export type CardanoOutput =
     | {
           addressParameters: CardanoAddressParameters,
@@ -151,12 +153,16 @@ export type CardanoCertificate = {
     path?: string | number[],
     pool?: string,
     poolParameters?: CardanoPoolParameters,
+    scriptHash?: string,
 };
 
 export type CardanoWithdrawal = {
-    path: string | number[],
+    path?: string | number[],
     amount: string,
+    scriptHash?: string,
 };
+
+export type CardanoMint = CardanoAssetGroup[];
 
 export type CardanoCatalystRegistrationParameters = {
     votingPublicKey: string,
@@ -178,9 +184,11 @@ export type CardanoSignTransaction = {
     certificates?: CardanoCertificate[],
     withdrawals?: CardanoWithdrawal[],
     validityIntervalStart?: string,
+    auxiliaryData?: CardanoAuxiliaryData,
+    mint?: CardanoMint,
+    additionalWitnessRequests?: (string | number[])[],
     protocolMagic: number,
     networkId: number,
-    auxiliaryData?: CardanoAuxiliaryData,
     signingMode: CardanoTxSigningMode,
 };
 
--- a/src/js/types/networks/cardano.js
+++ b/src/js/types/networks/cardano.js
@@ -4,9 +4,11 @@
 import type {
     HDNodeType,
     CardanoAddressType,
-    CardanoTxAuxiliaryDataSupplementType,
     CardanoCertificateType,
+    CardanoNativeScriptType,
+    CardanoNativeScriptHashDisplayFormat,
     CardanoPoolRelayType,
+    CardanoTxAuxiliaryDataSupplementType,
     CardanoTxSigningMode,
     CardanoTxWitnessType,
 } from '../trezor/protobuf';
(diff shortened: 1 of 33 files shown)


src/js/types/networks/cardano.js
```diff
@@ -67,7 +92,8 @@ export type CardanoInput = {
 
 export type CardanoToken = {
     assetNameBytes: string,
-    amount: string,
+    amount?: string,
+    mintAmount?: string,
```
mroz22, line 96: hello, some of the changes to this file are not reflected to `networks/cardano.d.ts` maybe you forgot to do so? For example in CardanoToken or CardanoAssetGroup

in general both typescript and flow declarations should be complete.

mroz22 (changes requested):
Hi, so far looks almost good, almost ready.

src/js/types/networks/cardano.js
davidmisiak: Right, sorry for that and thanks for noticing! `CardanoToken` fixed in https://github.com/trezor/connect/pull/920/commits/a7b2cd35a28f4e2e996cf2946c08a1a8e93fb24b. I went through both files and this seems to be the only difference - I haven't found a problem in `CardanoAssetGroup`, they both should be 
```
export type CardanoAssetGroup = {
    policyId: string,
    tokenAmounts: CardanoToken[],
};
```

mroz22:
Hi, thanks for resolving my comments. I think we are almost ready to merge it. There is just problem with running tests in browser env (karma). Could you please look at it? at the first glance, I couldn't figure out what might be wrong. 

I can run these tests locally in develop branch using

```
./tests/run.sh -s "yarn test:karma:production" -i getAddress
```

but trying to do the same in your branch fails in a really weird way

src/js/types/networks/cardano.js
mroz22: ok thanks.

davidmisiak:
> There is just problem with running tests in browser env (karma). Could you please look at it?

This should be fixed in https://github.com/trezor/connect/pull/920/commits/4e71afedd872d2d6b0d922978d9a2b403b3193bc (the wrong import path apparently prevented all Karma tests from running).

mroz22:
Thanks a lot for finding it. We should make it possible to import like you did in the beginning but this is out of the scope of this PR. 

Now please squash your fixups and rebase on latest develop. Then we are ready to go I think.

davidmisiak:
> Now please squash your fixups and rebase on latest develop. Then we are ready to go I think.

Done.

mroz22 (approved):
Ready to merge imho. Lets wait a while if @szymonlesisz wants to look at it also.

szymonlesisz:
looks good and working with older FW as well, thanks guys!

davidmisiak:
Hi, if my information is correct, there won't be a November release of Trezor firmware and possibly also Trezor Connect. However, we have already implemented multi-sig related changes to cardano-hw-cli (both for Trezor and Ledger) and we would like to release it with multisig support sooner than December (because of Ledger). It would be much simpler for us to include a current Connect build (with multi-sig support) in cardano-hw-cli and release it with a notice, that this release brings multisig functionality only to Ledger users. This way, we wouldn't have to revert our multi-sig changes to the Trezor part of cardano-hw-cli and we would only limit multi-sig functionality in cardano-hw-cli. The new Connect still works with 2.4.2 firmware (just the new functionality is missing), so everything should be fine in this regard. We would create an extended release in our fork of Trezor Connect and include that in cardano-hw-cli until an official release would be made. Would you have any issues with this approach please?

davidmisiak:
@mroz22 or @szymonlesisz would you have a moment to take a look at this please?

davidmisiak:
Or do you plan to release Connect sooner than in December (with Trezor Firmware)?

PR #926, opened 2021-10-30
fix: ethereumSignTransaction eip1559 support for T1

mmilata:
In trezor/trezor-firmware#1861 we've added [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) support to T1, and @bosomt found that it doesn't work with mycrypto wallet. It looks like it needs to be whitelisted here?


mroz22:
lets continue with #927

PR #933, opened 2021-11-09
Cardano derivation

mroz22:
In this PR I would like to implement trezor-firmware changes as proposed here https://github.com/trezor/trezor-firmware/pull/1873

So far I have only added the additional params `derive_cardano` and `derivation_type`. Work in progress. 

> If the user has BIP-39 seed, and Initialize(derive_cardano=True) is not sent, all Cardano calls will fail because the root secret will not be available. The only way to preserve compatibility here would be to use True as default -- but that would incur start-up delay for all other wallets, which is undesirable.

- I believe we need a default value to be sent from Connect as we can't reasonably expect all 3rd parties to change their implementations. 
- the question is what value shoudl we decide for. My initial idea was to send `derive_cardano=true` unless specified otherwise for some period of time and with a next release change this fallback to `derive_cardano=false`. This is the defensive strategy. If we are sure that there are no wallets that support cardano together with other coins, we can safely set the aformentioned fallback to `derive_cardano=false`. 
 
> The cardano_derivation argument is required on every Cardano call. We could set a default value here. But it would have to be Icarus-Trezor (otherwise the wallets would silently start using a different derivation for people with 24-word seeds). And we kind of want the default to be Icarus going forward --- especially for new users creating new accounts. (This is why trezorctl already defaults to Icarus.)  This way, we pass the "default value" question onto Connect. Personally I believe that Connect should also require the argument and pass the decision to individual wallets, but this is up for debate with the Connect team.

- I believe we need a default value to be sent from Connect as we can't reasonably expect all 3rd parties to change their implementations. 
- TODO: Icarus-Trezor?

### DIFF
--- a/src/js/core/methods/CardanoGetAddress.js
+++ b/src/js/core/methods/CardanoGetAddress.js
@@ -17,6 +17,7 @@ import type { CoreMessage } from '../../types';
 import type { CardanoAddress } from '../../types/networks/cardano';
 import { CardanoAddressType } from '../../types/networks/cardano';
 import type { CardanoAddressParametersType, MessageType } from '../../types/trezor/protobuf';
+import { Enum_CardanoDerivationType as CardanoDerivationType } from '../../types/trezor/protobuf';
 
 type Params = {
     ...$ElementType<MessageType, 'CardanoGetAddress'>,
@@ -60,6 +61,7 @@ export default class CardanoGetAddress extends AbstractMethod {
                 { name: 'addressParameters', type: 'object', obligatory: true },
                 { name: 'networkId', type: 'number', obligatory: true },
                 { name: 'protocolMagic', type: 'number', obligatory: true },
+                { name: 'derivationType', type: 'number' },
                 { name: 'address', type: 'string' },
                 { name: 'showOnTrezor', type: 'boolean' },
             ]);
@@ -76,6 +78,10 @@ export default class CardanoGetAddress extends AbstractMethod {
                 address: batch.address,
                 protocol_magic: batch.protocolMagic,
                 network_id: batch.networkId,
+                derivation_type:
+                    typeof batch.derivationType !== 'undefined'
+                        ? batch.derivationType
+                        : CardanoDerivationType.ICARUS_TREZOR,
                 show_display: showOnTrezor,
             });
         });
@@ -152,12 +158,19 @@ export default class CardanoGetAddress extends AbstractMethod {
         return uiResp.payload;
     }
 
-    async _call({ address_parameters, protocol_magic, network_id, show_display }: Params) {
+    async _call({
+        address_parameters,
+        protocol_magic,
+        network_id,
+        derivation_type,
+        show_display,
+    }: Params) {
         const cmd = this.device.getCommands();
         const response = await cmd.typedCall('CardanoGetAddress', 'CardanoAddress', {
             address_parameters,
             protocol_magic,
             network_id,
+            derivation_type,
             show_display,
         });
         return response.message;
--- a/src/js/core/methods/CardanoGetPublicKey.js
+++ b/src/js/core/methods/CardanoGetPublicKey.js
@@ -11,6 +11,7 @@ import { UiMessage } from '../../message/builder';
 import type { CoreMessage } from '../../types';
 import type { CardanoPublicKey } from '../../types/networks/cardano';
 import type { MessageType } from '../../types/trezor/protobuf';
+import { Enum_CardanoDerivationType as CardanoDerivationType } from '../../types/trezor/protobuf';
 
 export default class CardanoGetPublicKey extends AbstractMethod {
     params: $ElementType<MessageType, 'CardanoGetPublicKey'>[] = [];
@@ -43,6 +44,7 @@ export default class CardanoGetPublicKey extends AbstractMethod {
             // validate incoming parameters for each batch
             validateParams(batch, [
                 { name: 'path', obligatory: true },
+                { name: 'derivationType', type: 'number' },
                 { name: 'showOnTrezor', type: 'boolean' },
             ]);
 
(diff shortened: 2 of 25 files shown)


matejcik:
> * I believe we need a default value to be sent from Connect as we can't reasonably expect all 3rd parties to change their implementations.

You don't actually need to supply a value here. If you don't, it defaults to `false`.
Also if you send `Initialize(derive_cardano=X)` and then `Initialize(session_id=...)`, the value is still X for that session_id.

You can also send `Initialize()`, then `Initialize(session_id=..., derive_cardano=Y)` and unless the seed is already derived, the value will flip to Y for that session.

Meaning that the following strategy seems viable:
- by default, don't set `derive_cardano` (or send `derive_cardano=null` or w/e)
- if you're doing a Cardano call, send `derive_cardano=true` just beforehand
- wallets that do _NOT_ use Cardano will have a perfect user experience
- wallets that _ONLY_ use Cardano will also have a perfect user experience
- wallets that do both, AND are not updated to manage this parameter on their own, AND call something like `GetPublicKey` first and `CardanoGetPublicKey` second, will have degraded user experience - the user will be asked for the passphrase twice the first time around. After the first `Initialize(derive_cardano=true)`, and for as long as that session_id is alive, both Cardano and non-Cardano calls will work OK and not ask for a passphrase again.

> * send `derive_cardano=true` unless specified otherwise for some period of time

Just for completeness, i'll restate my argument here: I don't think burdening other wallets because of Cardano-specific 3rd parties is reasonable.
OTOH this is potentially breaking for Cardano wallets but only degrading for non-Cardano so :woman_shrugging: 

---

> * TODO: Icarus-Trezor?

The default cannot be anything other than `Icarus-Trezor` because otherwise we are **silently breaking** wallets for users with 24-word seeds, with (a) no indication about what the problem is besides "i recently updated firmware", and (b) no way to recover except to switch to a different wallet or something.

src/js/core/methods/CardanoGetPublicKey.js
```diff
@@ -43,6 +44,7 @@ export default class CardanoGetPublicKey extends AbstractMethod {
             // validate incoming parameters for each batch
             validateParams(batch, [
                 { name: 'path', obligatory: true },
+                { name: 'derivation_type' },
```
szymonlesisz:
```suggestion
               { name: 'derivation_type', type: 'number' },
```

src/js/device/DeviceCommands.js
```diff
@@ -477,29 +477,13 @@ export default class DeviceCommands {
...
-            default: {
-                const { message } = await this.typedCall('GetAddress', 'Address', {
-                    address_n: [toHardened(44), toHardened(1), toHardened(0), 0, 0],
-                    coin_name: 'Testnet',
-                    script_type: 'SPENDADDRESS',
-                });
-                return message.address;
-            }
-        }
+    async _getAddressForNetworkType() {
```
szymonlesisz: use previous/legacy code to fallback older FW

szymonlesisz (changes requested):
missing `CommonParams.useCardanoDerivation` in types/params (and ts)

mroz22:
we are soon about to merge this PR so just pinging @gabrielKerekes in case you want to test the new firmware and trezor-connect that will go live on December 8 in your implementation. I am not aware of any breaking changes though, so we should be safe.

refi93:
@mroz22 - can you clarify what would happen to Cardano wallets that don't update their integration at all? Will it still be working and if so will there possibly be some inefficiencies such as redundant prompts either in connect or the trezor device?

Also, if I understand correctly, wallets would pass some additional parameter when initializing the wallet but I don't see the docs being updated in the PR, but maybe I just missed that

gabrielKerekes:
Thanks @mroz22. So if I understand this correctly, for now third parties don't have to do anything. In some near future the default derivation type might be removed and third parties will need to provide it? And it's probably best they also start sending the `derive_cardano` common param?

mroz22:
@refi93 thanks for your notice about docs/readme. I totally forgot about that. Will push some update. Regarding your question - if your wallet works only with cardano and no other coins, you should be ok, nothing changes for you. If on the other hand you aim to provide multi-coin support you might need to make some changes, but this is not your case if I remember correctly?

mroz22:
@gabrielKerekes yes no need to hurry at the moment. And I don't think we will be removing these defaults in any foreseeable future. Just wanted to let you know that you already have such option - to specify derivation_type if you need to achieve better compatibility with another wallets. I understand that we need to have this better documented so I asked firmware people to write up some summary in documetnation in trezor-firmware repo.

src/js/core/methods/CardanoGetAddress.js
```diff
@@ -76,6 +78,7 @@ export default class CardanoGetAddress extends AbstractMethod {
                 address: batch.address,
                 protocol_magic: batch.protocolMagic,
                 network_id: batch.networkId,
+                derivation_type: batch.derivation_type || CardanoDerivationType.ICARUS_TREZOR,
```
davidmisiak: I think this is not correct - `CardanoDerivationType.LEDGER` has value `0`, so it gets replaced by `CardanoDerivationType.ICARUS_TREZOR`. I tested it (with the mnemonic from [firmware derivation tests](https://github.com/vacuumlabs/trezor-firmware/blob/master/common/tests/fixtures/cardano/get_base_address.derivations.json)) and indeed `LEDGER` gives the same signature as `ICARUS_TREZOR` (which is different from the `ICARUS` one).

This line is present multiple times below.

mroz22: !!! great observation. thanks. I will fix that and also add some tests to cover this.

davidmisiak:
> pinging @gabrielKerekes in case you want to test the new firmware and trezor-connect that will go live on December 8 in your implementation

We tested this with cardano-hw-cli and everything seems to work. Thanks for pinging us!

refi93:
> but this is not your case if I remember correctly?

@mroz22 actually, we are working on a multicoin wallet as well (not publicly released yet), though the only trezor-supported coin is Cardano. So if you could clarify what changes are needed for multicoin wallets, that would be great

szymonlesisz (approved):

PR #945, opened 2021-11-17
fix: new DOGE fee policy

szymonlesisz:
https://github.com/dogecoin/dogecoin/blob/master/doc/fee-recommendation.md#dust-limits

- default fee 1000 (0.01) DOGE
- min fee 100 (0.001) DOGE
- dust limit 1000 (0.01) DOGE

- round/floor fee is no longer neccessary
- base fee (1 DOGE) no longer necessary

Those values are currently tested by QA

Related trezor-firmware PR https://github.com/trezor/trezor-firmware/pull/1921

### DIFF
--- a/src/data/coins.json
+++ b/src/data/coins.json
@@ -914,9 +914,9 @@
             "decimals": 8,
             "decred": false,
             "default_fee_b": {
-                "Normal": 100000
+                "Normal": 1000
             },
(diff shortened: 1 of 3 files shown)


marekrjpolak (approved):
:shipit:

mroz22 (approved):

PR #948, opened 2021-11-19
tests(cardano): run tests for derivation types

mroz22:
Ok, I wanted to add some more tests on cardano derivation types. But I found a thing instead!

In this configuration, tests fail with a weird bridge bug `device disconnected during action` but this does not seem to be the case. emulator is still alive.

This tests works fine when run against bridge `version: 2.0.27`.

So I am leaving it here as a working example of bridge buggy behavior. Hope I will find enought inner strenght to do the debugging part also :wink:


mroz22:
another case where the latest bridge is displaying buggy behavior but older bridge works is in trezor/suite connect-popup integration tests. 
https://github.com/trezor/trezor-suite/pull/4581

todo: close this issue here and move it into trezor-suite so that it gets more attention

PR #952, opened 2021-11-22
tests: use latest released version as default

mroz22:
hmm :thinking: this is a double edged sword.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -76,8 +76,13 @@ show_usage() {
   echo "  -s       actual test script. default: 'yarn test:integration'"
 }
 
+# Find the latest released firmware version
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -76,8 +76,13 @@ show_usage() {
   echo "  -s       actual test script. default: 'yarn test:integration'"
 }
 
+# Find the latest released firmware version
+RELEASED_FIRMWARE=$(curl https://raw.githubusercontent.com/trezor/webwallet-data/master/firmware/2/releases.json | tac | tac | grep -m1 -o -P '(?<="version\": \[).*(?=\])' | sed 's/, /./g')
```
vdovhanych, line 80: won't this still fail if we runt the script in the ci ?

mroz22: it wokrs in github, so I would say that as long as configuration of the base image is ok, ti should pass

vdovhanych (approved):

PR #930, opened 2021-11-05
Stellar: add support for StellarManageBuyOfferOp and StellarPathPaymentStrictSendOp.

overcat:
continuation of https://github.com/trezor/trezor-firmware/pull/1838
- updated trezor-common
- rename StellarManageOfferOp to StellarManageSellOfferOp
- rename StellarPathPaymentOp to StellarPathPaymentStrictReceiveOp
- rename StellarCreatePassiveOfferOp to StellarCreatePassiveSellOfferOp
- add support for StellarManageBuyOfferOp
- add support for StellarPathPaymentStrictSendOp

In c933d23079047ec9b97e3213a0265c6e5870d777, I only submitted changes related to Stellar.

I hope I didn't miss anything else, please let me know if there is any.

PR #953, opened 2021-11-23
tests: stellar 2.4.2 fixtures and _ensureFeatureIsSupported

mroz22:
hello @overcat 

I have just realized that after merging your PR https://github.com/trezor/connect/pull/930 we have problem to run tests against 2.4.2 firmware (which is currently the latest released firmware)

`./tests/run.sh -i stellarSignTransaction -f 2.4.2`

There are 2 problems: 
- unknown messages in older firmwares (StellarManageBuyOfferOp, StellarPathPaymentStrictSendOp). This might be fixed by adding something like this (inspiration from CardanoSign.tx). Basically when you detect that params contain some operation unknown to an older firmware, you may raise some more descriptive error than what is otherwise received from device. 
https://github.com/trezor/connect/blob/b23d81af875173b0f732e17d7ddf51e8a399962a/src/js/core/methods/CardanoSignTransaction.js#L209-L215

- the other problem is that fixture `timebounds-0-0` returns different signature. I would appreciate your input here.

### DIFF
--- a/src/js/core/methods/StellarSignTransaction.js
+++ b/src/js/core/methods/StellarSignTransaction.js
@@ -5,6 +5,7 @@ import { validateParams, getFirmwareRange } from './helpers/paramsValidator';
 import { getMiscNetwork } from '../../data/CoinInfo';
 import { validatePath } from '../../utils/pathUtils';
 import * as helper from './helpers/stellarSignTx';
+import { ERRORS } from '../../constants';
 
 import type { CoreMessage } from '../../types';
 
@@ -13,6 +14,11 @@ type Params = {
     networkPassphrase: string,
     transaction: any,
 };
+
+const StellarSignTransactionFeatures = Object.freeze({
+    manageBuyOffer: ['1.10.4', '2.4.3'],
+    pathPaymentStrictSend: ['1.10.4', '2.4.3'],
+});
 
 export default class StellarSignTransaction extends AbstractMethod {
     params: Params;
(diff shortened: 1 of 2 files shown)


overcat:
Hi, @mroz22 

> * unknown messages in older firmwares (StellarManageBuyOfferOp, StellarPathPaymentStrictSendOp). This might be fixed by adding something like this (inspiration from CardanoSign.tx). Basically when you detect that params contain some operation unknown to an older firmware, you may raise some more descriptive error than what is otherwise received from device.

I believe this is a great implement

> * the other problem is that fixture `timebounds-0-0` returns different signature. I would appreciate your input here.

I think this may be caused by PR[#1755](https://github.com/trezor/trezor-firmware/pull/1755), can you confirm it?

overcat:
Hello @mroz22, I have confirmed that the second problem is caused by PR[#1755](https://github.com/trezor/trezor-firmware/pull/1755), the different handling of timebounds in 2.4.2 and 2.4.3 caused this difference.

mroz22:
Good, thanks @overcat 
And do you know whether both signatures are valid? Or do we need to take some action, for example forbid stellarSignTransaction method for older firmwares. Or maybe forbid it for older firmwares if call params contain zero timebounds?

overcat:
For version 2.4.2, if zero timebounds is included, the signature generated is invalid and cannot be accepted by the stellar network. The signature generated by version 2.4.3 can be accepted.

mroz22:
ok, so from my perspective, I am OK with it as it is. Feel free to add some additional validation of input params if you feel like you want to improve user experience. 

At the moment we need to fix trezor-user-env repository which is breaking tests and after that we can merge this PR.

overcat:
Thanks, I will consider adding it in another PR.

src/js/core/methods/StellarSignTransaction.js
```diff
@@ -14,6 +15,11 @@ type Params = {
     transaction: any,
 };
 
+const StellarSignTransactionFeatures = Object.freeze({
+    manageBuyOffer: ['0', '2.4.3'],
```
overcat: Does the first value in the list refer to the version number of Model One firmware? If it is, I think it should be set to 1.10.4

mroz22: thanks. fixed.

overcat: Another question is, do we only need to check two operations? What about operations such as StellarCreatePassiveSellOfferOp?

I provide a complete list here.
https://github.com/trezor/connect/pull/954/files

marekrjpolak (approved):
LGTM

src/js/core/methods/StellarSignTransaction.js
mroz22: Feel free to investigate and add further improvements. If the other operations are not checked - user will get a less descriptive error generated by trezor-firmware. What we are doing here now is more "nice to have" than something really neccessary. Pull requests are always welcome :wink:

overcat: Get it, thanks!

PR #954, opened 2021-11-24
feat(stellar): ensure the parameter is supported by the hardware.

overcat:


overcat:
Hello @mroz22, I did not notice that you have implemented this feature in #953, so I think I can just close this PR.

src/js/core/methods/StellarSignTransaction.js
```diff
@@ -6,12 +6,29 @@ import { getMiscNetwork } from '../../data/CoinInfo';
 import { validatePath } from '../../utils/pathUtils';
 import * as helper from './helpers/stellarSignTx';
 
-import type { CoreMessage } from '../../types';
+import type { CoreMessage, StellarTransaction } from '../../types';
+import { ERRORS } from '../../constants';
+
+const StellarSignTransactionFeatures = Object.freeze({
+    StellarCreateAccountOp: ['1.10.3', '2.4.2'],
```
overcat, line 13: Is it okay to force users to upgrade to ^1.10.3 and ^2.4.2? If not, we need to check in which version these operations were added, and then fix it.

mroz22: This is not really forcing. It just tells users that want to use these exact params to update. Which is better then present them with message "unexpected message error". So without updating this kind of call will not work for them anyway. 

If you will, feel free to investigate at which point these operations were added and open another PR.

PR #956, opened 2021-11-29
fix: missing blockchain link for bitcoin regtest

karliatto:
This PR adds missing blockchain link `type` and `url` for Bitcoin regtest.

### DIFF
--- a/src/data/coins.json
+++ b/src/data/coins.json
@@ -62,7 +62,12 @@
             "address_type": 111,
             "address_type_p2sh": 196,
             "bech32_prefix": "bcrt",
-            "blockchain_link": null,
+            "blockchain_link": {
+                "type": "blockbook",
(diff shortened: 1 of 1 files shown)


szymonlesisz:
i think this should be handled on trezor-suite side (using settings/crypto/advanced)
following [the docs](https://github.com/trezor/connect/blob/develop/docs/supported-coins.md) the source of coins.json file is in trezor-firmware repo and teoretically this change should be added there **but** i dont think it needed at all. see my comment in the [commit](https://github.com/trezor/trezor-suite/pull/4540/commits/b3b7d96cd24ee17c6f3b7556500f3920db855b86) that you already cherry picked in your branch.

karliatto:
Closing it, since it does not make sense to have those changes here.

PR #968, opened 2021-12-10
fix: trezor-link set fetch for node env

mroz22:


yarn.lock
```diff
@@ -3638,7 +3638,7 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4, create-hmac@^1.1.7:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-cross-fetch@^3.0.6:
+cross-fetch@^3.0.6, cross-fetch@^3.1.4:
```
matejkriz, line 3641: maybe we could also bump cross-fetch version in @trezor/rollout so we don't need two versions here?

matejkriz (approved):

yarn.lock
mroz22: cool idea.

PR #969, opened 2021-12-10
fix(cardano): fix catalyst reward address backwards compatibility

gabrielKerekes:
A bug in Cardano transaction signing was reported to us yesterday evening. Trezor FW 2.4.3 introduced a breaking change in Cardano reward addresses. It expects the reward address path to be sent in the `stakingPath` field instead of the `path` field. Trezor Connect should be handling this for the `cardanoGetAddress` call but we forgot to include backwards compatibility for Catalyst registration which also uses reward addresses. This PR addresses that i.e. catalyst address parameters will also get transformed to be compatible with older FWs.

Given that Catalyst registration is currently broken, it would be great if we could get at least a beta release out ASAP.

### DIFF
--- a/docs/methods/cardanoSignTransaction.md
+++ b/docs/methods/cardanoSignTransaction.md
@@ -259,7 +259,7 @@ TrezorConnect.cardanoSignTransaction({
             stakingPath: "m/1852'/1815'/0'/2/0",
             rewardAddressParameters: {
                 addressType: CardanoAddressType.REWARD,
-                path: "m/1852'/1815'/0'/2/0",
+                stakingPath: "m/1852'/1815'/0'/2/0",
             },
             nonce: "22634813",
         },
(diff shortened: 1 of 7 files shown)


mroz22:
Thanks. I think there will be release on Monday with this included.

mroz22 (approved):

gabrielKerekes:
> I think there will be release on Monday with this included.

That would be awesome. Please let us know after it's released.

prusnak:
@gabrielKerekes We see reports like these since yesterday's update: https://forum.trezor.io/t/change-address-payment-credential-is-a-path/5345/3

Do you see it as well on your side?

gabrielKerekes:
This shouldn't be related to yesterday's update, but to the FW update, specifically to the Cardano multi-sig update https://github.com/trezor/trezor-firmware/pull/1772.

The change in the UI is mentioned in the PR description:
> - Because of the new address types the `get-address` and `sign_tx` change output UI flow would become too complicated. The UI flow was thus refactored and changed a bit - hopefully simplifying it. [You can take a look at the designs here.](https://www.figma.com/file/x2MhF1Jb99AfBzv0Uc34gb/Trezor-address-derivation-UI?node-id=0%3A1)

However, users should in general not see that message because when the address credentials match, then the change output is hidden and I think they should always match in standard wallets e.g. Adalite.

prusnak:
Thanks for the explanation Gabriel. It seems the bug is present only in the Yoroi wallet.

PR #971, opened 2021-12-11
fix(stellar): fix transformTransaction

overcat:
`manageBuyOffer.buyAmount` is equivalent to `StellarManageBuyOfferOp.amount`, we need to convert here.
See https://github.com/stellar/js-stellar-base/blob/master/src/operations/manage_buy_offer.js#L12

---

fee is string type
See https://github.com/stellar/js-stellar-base/blob/master/src/transaction.js#L55

---

text memo can be String or Buffer in SDK

### DIFF
--- a/src/js/plugins/stellar/plugin.js
+++ b/src/js/plugins/stellar/plugin.js
@@ -163,7 +163,7 @@ const transformTransaction = (path, transaction) => {
         networkPassphrase: transaction.networkPassphrase,
         transaction: {
             source: transaction.source,
-            fee: transaction.fee,
+            fee: Number.parseInt(transaction.fee, 10),
             sequence: transaction.sequence,
             memo: transformMemo(transaction.memo),
             timebounds: transformTimebounds(transaction.timeBounds),
--- a/src/js/plugins/stellar/plugin.js
+++ b/src/js/plugins/stellar/plugin.js
@@ -64,7 +64,7 @@ const transformAmount = amount => new BigNumber(amount).times(10000000).toString
 const transformMemo = memo => {
     switch (memo.type) {
         case StellarSdk.MemoText:
-            return { type: 1, text: memo.value };
+            return { type: 1, text: memo.value.toString('utf-8') };
         case StellarSdk.MemoID:
             return { type: 2, id: memo.value };
         case StellarSdk.MemoHash:
--- a/src/js/plugins/stellar/plugin.js
+++ b/src/js/plugins/stellar/plugin.js
@@ -100,7 +100,7 @@ const transformTimebounds = timebounds => {
  * @returns {TrezorConnect.StellarTransaction}
  */
 const transformTransaction = (path, transaction) => {
-    const amounts = ['amount', 'sendMax', 'destAmount', 'startingBalance', 'limit'];
+    const amounts = ['amount', 'sendMax', 'destAmount', 'startingBalance', 'limit', 'buyAmount'];
     const assets = ['asset', 'sendAsset', 'destAsset', 'selling', 'buying', 'line'];
 
     const operations = transaction.operations.map((o, i) => {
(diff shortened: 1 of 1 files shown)


mroz22 (approved):
looks good. I believe we are going to release this altogether with other fixes within 24 hours.

overcat:
Hi @mroz22, I updated this PR, added some fixes

src/js/plugins/stellar/plugin.js
```diff
@@ -64,7 +64,7 @@ const transformAmount = amount => new BigNumber(amount).times(10000000).toString
 const transformMemo = memo => {
     switch (memo.type) {
         case StellarSdk.MemoText:
-            return { type: 1, text: memo.value };
+            return { type: 1, text: memo.value.toString('utf-8') };
```
mroz22, line 67: could you please past link to docs about memo type?

overcat: https://github.com/stellar/js-stellar-base/blob/master/types/index.d.ts#L204

overcat:
Another question, is `const operation = Object.assign({}, o);` equivalent to `const operation = { ...o };`? I noticed that other projects use the former

mroz22:
> Another question, is `const operation = Object.assign({}, o);` equivalent to `const operation = { ...o };`? I noticed that other projects use the former

This is true. There is some internal discussion on this. The thing is that using spread operator `...` produces much larger javascript code after transpilation, so we actually don't mind using less sexy way o writing the same which is Object.assign.

src/js/plugins/stellar/plugin.js
mroz22: and can't this happen to be `string` type? in that case toString method call would fail with 'utf-8' argument.

overcat: I'm not sure what you are referring to, but hope this helps

https://github.com/trezor/trezor-firmware/pull/1745#discussion_r683263945

> In XDR the memo text is defined as string, which per XDR RFC must be ASCII, per Stellar docs probably UTF. In any case what you get from the parsed XDR should be of type str, right? If the XDR parser returns bytes, then the decode() call is correct, but ... if the decoding fails we need to raise an exception anyway because the incoming XDR is incorrect. So maybe remove the whole try/except block and let the original UnicodeDecodeError fall through?

overcat:
```
const StellarSDK = require("stellar-sdk")
// text memo
// memo value: "hello"
let tx = StellarSDK.TransactionBuilder.fromXDR("AAAAAgAAAADQ7sFU88KfOExfripsgTcWhrq3exqUlYC1x5Qphkv+kwAAF3AAAACNbLKV6gAAAAEAAAAAAAAAAAAAAABjlFcZAAAAAQAAAAVoZWxsbwAAAAAAAAEAAAAAAAAABAAAAAAAAAABVVNEQwAAAAA7mRE4Dv6Yi6CokA6xz+RPNm99vpRr7QdyQPf2JN8VxQAAAAABMS0AAAAAAQAAAAEAAAAAAAAAAA==", StellarSDK.Networks.PUBLIC)
console.log(tx.memo.value instanceof Buffer) // true
console.log(tx.memo.value.toString("utf-8"))
```

mroz22: Yep all good. Sorry I got a bit confused. toString method works in the same way for both String and Buffer here.

mroz22 (approved):

overcat:
Rebased.

mroz22:
Will ask you for one more rebase after this is merged #974  should happen in a while.

mroz22:
So https://github.com/trezor/connect/pull/974 is merge so please one more rebase :pray:

overcat:
> So #974 is merge so please one more rebase 🙏

rebased

PR #972, opened 2021-12-11
docs(Core): increase delay after RebootToBootloader and fix comment

tsusanka:
See this for rationale https://github.com/trezor/trezor-firmware/issues/1983. There is also a suggestion from @hiviah:

> 501 ms is minimum, 1 s should be OK. Although waiting 501 ms, then doing let's say 5 retry calls to GetFeatures until you get response instead of libusb error spaced 100 ms between each retry would make it slightly faster.

Which would be nice but I think we can leave it with 1s for now.

### DIFF
--- a/src/js/core/Core.js
+++ b/src/js/core/Core.js
@@ -657,9 +657,9 @@ export const onCall = async (message: CoreMessage) => {
 
         if (response) {
             if (method.name === 'rebootToBootloader' && response.success) {
-                // trezord may not detect auto reboot
-                // wait for device to switch to bootloader
-                await resolveAfter(501);
+                // Wait for device to switch to bootloader
+                // This delay is crucial see https://github.com/trezor/trezor-firmware/issues/1983
+                await resolveAfter(1000);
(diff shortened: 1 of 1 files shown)


src/js/core/Core.js
```diff
@@ -657,9 +657,9 @@ export const onCall = async (message: CoreMessage) => {
 
         if (response) {
             if (method.name === 'rebootToBootloader' && response.success) {
-                // trezord may not detect auto reboot
-                // wait for device to switch to bootloader
-                await resolveAfter(501);
+                // Wait for device to switch to bootloader
+                // This delay is crucial see trezor-firmware#1983
```
prusnak: Let's use the full link

```suggestion
                // This delay is crucial see https://github.com/trezor/trezor-firmware/issues/1983
```

prusnak (changes requested):

prusnak (approved):

szymonlesisz (approved):

PR #979, opened 2021-12-14
chore: remove unused methods

### LINKED ISSUE #977: Code cleanup
method not used `getCoinInfoByCapability`

mroz22:
close #977

### DIFF
--- a/src/js/data/CoinInfo.js
+++ b/src/js/data/CoinInfo.js
@@ -338,30 +338,3 @@ export const getUniqueNetworks = (networks: Array<?CoinInfo>): CoinInfo[] =>
 
 export const getAllNetworks = (): CoinInfo[] =>
     [].concat(bitcoinNetworks).concat(ethereumNetworks).concat(miscNetworks);
-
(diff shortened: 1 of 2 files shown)


src/js/data/CoinInfo.js
```diff
@@ -110,53 +109,6 @@ export const fixCoinInfoNetwork = (ci: BitcoinNetworkInfo, path: number[]) => {
...
-    if (data.subversion.startsWith('/Bitcoin Cash')) {
-        return 'bch';
-    }
-    if (data.subversion.startsWith('/Bitcoin Gold')) {
-        return 'btg';
-    }
-    return 'btc';
-};
-
-export const getCoinInfoByHash = (hash: string, networkInfo: any) => {
```
szymonlesisz: i guess this should be implemented not removed, afaik this was designed to validate backend connection by 1st blockhash (data source from trezor-common)
it should be used every time you connect to a backend which may be malicious, it may claim that the shortcut is "Testnet" but in fact it will be "Btc"

szymonlesisz (approved):

Issue #825, opened 2021-05-31
Ethereum EIP-712 structured signatures isn't supported


artall64:
Reference: https://eips.ethereum.org/EIPS/eip-712
This EIP become widely use over the Ethereum Network, but looks like it's isn't available via SDK

Any plans to support this ?

szymonlesisz:
depends on https://github.com/trezor/trezor-firmware/pull/1568

ghost:
> Reference: https://eips.ethereum.org/EIPS/eip-712
> This EIP become widely use over the Ethereum Network, but looks like it's isn't available via SDK
> 
> Any plans to support this ?

You can ask them on this thread over here https://github.com/trezor/trezor-firmware/pull/1568

aydnep:
Hi guys. any updates on EIP-712 support?

ghost:
> Hi guys. any updates on EIP-712 support?

It looks like it's being tracked over here https://github.com/trezor/trezor-firmware/pull/1835

yura-pakhuchiy:
Any updates on this? trezor/trezor-firmware#1835 is merged, but not really usable without connect counterpart

yura-pakhuchiy:
@xlab maybe you have some code for connect ready?

mroz22:
Hello there. We don't have it in our plans in near future but we are happy to review and merge any contribution from comunity.

artall64:
Does't mean that EIP-712 support wouldn't be released on Trezor in 2021 ?
It's a key requirement for limit orders and much more DeFi stuff on Ethereum-like networks.

prusnak:
> Does't mean that EIP-712 support wouldn't be released on Trezor in 2021 ?

Will be released in early 2022. Please be patient.

tsusanka:
We have released EIP-712 support [in January](https://blog.trezor.io/trezor-suite-and-trezor-model-one-firmware-updates-january-2022-4a77e4a07a5a) release.

PR #982, opened 2021-12-15
feat(ethereum): EIP-712 signTypedData for Trezor T

matejcik:
Companion to https://github.com/trezor/trezor-firmware/pull/1835

~~Does not work at the moment! Tests are failing: `TypeError: TrezorConnect[testCase.method] is not a function`
It seems I'm failing to register the new method somehow?~~

~~Also type checking fails, and who knows how many bugs lurk -- not me for whom the code won't even execute.~~

PR #983, opened 2021-12-16
feat: Support Ethereum EIP-712 Sign Typed Data for Trezor T

aloisklink:
Adds support for EIP-712 Typed Signed Data #825.

This is mostly adapted from the Python implementation in `python/src/trezorlib/ethereum.py` in  https://github.com/trezor/trezor-firmware/pull/1835

~One potential issue is [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder), which is what I'm using to convert a UTF-8 string to bytes. It's supported in all the browsers Trezor Connect targets, **except IE**, and I'm unsure if Babel/Webpack is polyfilling it automatically. I've only run `karma` tests on Chrome.~ Replaced with `Buffer.from('utf-8').toString('hex')` in 0103d17334c81e9ba7e1b7b3ed39c7b396f35ba7

I also found quite a few Flow issues that I'm pretty sure would be fixed by updating Flow, (e.g. `await Promise` not always resolving the promise type) but unfortunately our [`.flowconfig` has deprecated options](https://github.com/trezor/connect/blob/240d017c9744792dd9747e73b9d34eb4e68bddca/.flowconfig#L16-L19) that aren't supported in the latest versions, so it's not trivial to update.

### DIFF
--- a/src/data/config.json
+++ b/src/data/config.json
@@ -162,6 +162,11 @@
             "methods": ["signTransaction"],
             "min": ["1.10.1", "2.4.0"],
             "comment": [""]
+        },
+        {
+            "methods": ["ethereumSignTypedData"],
+            "min": ["0", "2.4.3"],
+            "comment": ["EIP-712 typed signing support added in 2.4.3"]
         }
     ]
 }
--- a/src/js/core/methods/EthereumSignTypedData.js
+++ b/src/js/core/methods/EthereumSignTypedData.js
@@ -1,0 +1,163 @@
+/* @flow */
+
+import AbstractMethod from './AbstractMethod';
+import { validateParams, getFirmwareRange } from './helpers/paramsValidator';
+import { validatePath } from '../../utils/pathUtils';
+import { getEthereumNetwork } from '../../data/CoinInfo';
+import { toChecksumAddress, getNetworkLabel } from '../../utils/ethereumUtils';
+import type { CoreMessage, EthereumNetworkInfo } from '../../types';
+import type { MessageResponse, EthereumTypedDataStructAck } from '../../types/trezor/protobuf';
+import { ERRORS } from '../../constants';
+import type { EthereumSignTypedData as EthereumSignTypedDataParams } from '../../types/networks/ethereum';
+import { getFieldType, parseArrayType, encodeData } from './helpers/ethereumSignTypedData';
+
+type Params = {
+    ...EthereumSignTypedDataParams,
+    path: number[],
+    network?: EthereumNetworkInfo,
+};
+
+export default class EthereumSignTypedData extends AbstractMethod {
+    params: Params;
+
+    constructor(message: CoreMessage) {
+        super(message);
+
+        this.requiredPermissions = ['read', 'write'];
+
+        const { payload } = message;
+
+        // validate incoming parameters
+        validateParams(payload, [
+            { name: 'path', obligatory: true },
+            { name: 'data', type: 'object', obligatory: true },
+            { name: 'metamask_v4_compat', type: 'boolean', obligatory: true },
+        ]);
+
+        const path = validatePath(payload.path, 3);
+        const network = getEthereumNetwork(path);
+        this.firmwareRange = getFirmwareRange(this.name, network, this.firmwareRange);
+
+        this.info = getNetworkLabel('Sign #NETWORK typed data', network);
+
+        const { data, metamask_v4_compat } = payload;
+
+        this.params = {
+            path,
+            network,
+            data,
+            metamask_v4_compat,
+        };
+    }
+
+    async run() {
+        const cmd = this.device.getCommands();
+        const { path: address_n, network, data, metamask_v4_compat } = this.params;
+
+        const { types, primaryType, domain, message } = data;
+
+        let response: MessageResponse<
+            | 'EthereumTypedDataStructRequest'
+            | 'EthereumTypedDataValueRequest'
+            | 'EthereumTypedDataSignature',
+        > = await cmd.typedCall(
+            'EthereumSignTypedData',
+            // $FlowIssue typedCall problem with unions in response, TODO: accept unions
+            'EthereumTypedDataStructRequest|EthereumTypedDataValueRequest|EthereumTypedDataSignature',
+            {
+                address_n,
+                primary_type: primaryType,
+                metamask_v4_compat,
+            },
+        );
+
+        // sending all the type data
+        while (response.type === 'EthereumTypedDataStructRequest') {
+            // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+            const { name: typeDefinitionName } = response.message;
+            const typeDefinition = types[typeDefinitionName];
+            if (typeDefinition === undefined) {
+                throw ERRORS.TypedError(
+                    'Runtime',
+                    `Type ${typeDefinitionName} was not defined in types object`,
+                );
+            }
+            const dataStruckAck: EthereumTypedDataStructAck = {
+                members: typeDefinition.map(({ name, type: typeName }) => ({
+                    name,
+                    type: getFieldType(typeName, types),
+                })),
+            };
+            response = await cmd.typedCall(
+                'EthereumTypedDataStructAck',
+                // $FlowIssue typedCall problem with unions in response, TODO: accept unions
+                'EthereumTypedDataStructRequest|EthereumTypedDataValueRequest|EthereumTypedDataSignature',
+                dataStruckAck,
+            );
+        }
+
+        // sending the whole message to be signed
+        while (response.type === 'EthereumTypedDataValueRequest') {
+            // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+            const { member_path } = response.message;
+
+            let memberData;
+            let memberTypeName;
+
+            const [rootIndex, ...nestedMemberPath] = member_path;
+            switch (rootIndex) {
+                case 0:
+                    memberData = domain;
+                    memberTypeName = 'EIP712Domain';
+                    break;
+                case 1:
+                    memberData = message;
+                    memberTypeName = primaryType;
+                    break;
+                default:
+                    throw ERRORS.TypedError('Runtime', 'Root index can only be 0 or 1');
+            }
+
+            // It can be asking for a nested structure (the member path being [X, Y, Z, ...])
+            for (const index of nestedMemberPath) {
+                if (Array.isArray(memberData)) {
+                    memberTypeName = parseArrayType(memberTypeName).entryTypeName;
+                    memberData = memberData[index];
+                } else if (typeof memberData === 'object' && memberData !== null) {
+                    const memberTypeDefinition = types[memberTypeName][index];
+                    memberTypeName = memberTypeDefinition.type;
+                    memberData = memberData[memberTypeDefinition.name];
+                } else {
+                    // TODO: what to do when the value is missing (for example in recursive types)?
+                }
+            }
+
+            let encodedData;
+            // If we were asked for a list, first sending its length and we will be receiving
+            // requests for individual elements later
+            if (Array.isArray(memberData)) {
+                // Sending the length as uint16
+                encodedData = encodeData('uint16', memberData.length);
+            } else {
+                encodedData = encodeData(memberTypeName, memberData);
+            }
+
+            // $FlowIssue with `await` and Promises: https://github.com/facebook/flow/issues/5294, TODO: Update flow
+            response = await cmd.typedCall(
+                'EthereumTypedDataValueAck',
+                // $FlowIssue typedCall problem with unions in response, TODO: accept unions
+                'EthereumTypedDataValueRequest|EthereumTypedDataSignature',
+                {
+                    value: encodedData,
+                },
+            );
+        }
+
+        // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+        const { address, signature } = response.message;
+        return {
+            address: toChecksumAddress(address, network),
+            signature: `0x${signature}`,
+        };
+    }
+}
--- a/src/js/core/methods/helpers/ethereumSignTypedData.js
+++ b/src/js/core/methods/helpers/ethereumSignTypedData.js
@@ -1,0 +1,190 @@
+/* @flow */
+
+import BigNumber from 'bignumber.js';
+import type { EthereumSignTypedDataTypes } from '../../../types/networks/ethereum';
+import type { EthereumFieldType } from '../../../types/trezor/protobuf';
+import { Enum_EthereumDataType } from '../../../types/trezor/protobuf';
+import { ERRORS } from '../../../constants';
+import { stripHexPrefix } from '../../../utils/formatUtils';
+
+// Copied from https://github.com/ethers-io/ethers.js/blob/v5.5.2/packages/abi/src.ts/fragments.ts#L249
+const paramTypeArray = new RegExp(/^(.*)\[([0-9]*)\]$/);
+const paramTypeBytes = new RegExp(/^bytes([0-9]*)$/);
+const paramTypeNumber = new RegExp(/^(u?int)([0-9]*)$/);
+
+/**
+ * Parse the given EIP-712 array type into its entries, and its length (if not dynamic)
+ * E.g. `uint16[32]` will return `{entryTypeName: 'uint16', arraySize: 32}`.
+ */
+export function parseArrayType(arrayTypeName: string): {
+    entryTypeName: string,
+    arraySize: number | null,
+} {
+    const arrayMatch = paramTypeArray.exec(arrayTypeName);
+    if (arrayMatch === null) {
+        throw ERRORS.TypedError(
+            'Runtime',
+            `typename ${arrayTypeName} could not be parsed as an EIP-712 array`,
+        );
+    }
+    const [_, entryTypeName, arraySize] = arrayMatch;
+    return {
+        entryTypeName,
+        arraySize: parseInt(arraySize, 10) || null,
+    };
+}
+
+/**
+ * Converts a number to a two's complement representation.
+ *
+ * E.g. -128 would be 0x80 in two's complement, while 127 would be 0x7F.
+ *
+ * BigNumber.js has no built-in function, unlike https://www.npmjs.com/package/bn.js
+ */
+function twosComplement(number: BigNumber, bytes: number): BigNumber {
+    if (bytes < 1 || bytes > 32) {
+        throw ERRORS.TypedError(
+            'Runtime',
+            'Int byte size must be between 1 and 32 (8 and 256 bits)',
+        );
+    }
+    // Determine value range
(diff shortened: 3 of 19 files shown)


mroz22:
Hello, thanks for contribution! We are looking into it now.

aloisklink:
Ah, I'm sorry, I forgot to run an `yarn run lint` on my final test changes! I've added a `fixup!` commit that fixes those style issues.

src/data/config.json
```diff
@@ -162,6 +162,11 @@
             "methods": ["signTransaction"],
             "min": ["1.10.1", "2.4.0"],
             "comment": [""]
+        },
+        {
+            "methods": ["ethereumSignTypedData"],
+            "min": ["0", "2.4.2"],
+            "comment": ["EIP-712 typed signing support added in 2.4.2"]
```
matejcik: it should be 2.4.3

aloisklink: Good catch!

I've updated it. I also made sure sure to stick `(Trezor T only)` into the CHANGELOG.md entry.

aloisklink:
Hi @matejcik, sorry, I didn't see your PR #982 that does the same thing.

Feel free to close this if you'd prefer to go with your implementation. I'm happy for you to copy-and-paste any docs/type entries from this PR to your PR if you decide to merge your one instead.

src/js/core/methods/helpers/ethereumTypeDataConversions.js
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Converts a number to a two's compliment representation.
+ *
+ * E.g. -128 would be 0x80 in two's compliment, while 127 would be 0x7F.
+ *
+ * BigNumber.js has no built-in function, unlike https://www.npmjs.com/package/bn.js
+ */
+function twosCompliment(number: BigNumber, bytes: number): BigNumber {
```
matejcik: typo: compl**e**ment
```suggestion
function twosComplement(number: BigNumber, bytes: number): BigNumber {
```

src/js/core/methods/helpers/ethereumTypeDataConversions.js
```diff
@@ -0,0 +1,178 @@
...
+ * @param typeName - Primitive Solidity data type (e.g. `uint16`)
+ * @param data - The actual data to convert.
+ * @returns Hex string of the data.
+ */
+export function encodeData(typeName: string, data: any): string | Uint8Array {
+    if (paramTypeBytes.test(typeName) || typeName === 'address') {
+        return stripHexPrefix(data);
+    }
+    if (typeName === 'string') {
+        return new TextEncoder().encode(data);
```
matejcik: the codebase seems to use `Buffer.from('utf-8').toString('hex')`

src/js/core/methods/helpers/ethereumTypeDataConversions.js
```diff
@@ -0,0 +1,178 @@
...
+export function getFieldType(
+    typeName: string,
+    types: EthereumSignTypedDataTypes,
+): EthereumFieldType {
+    const arrayMatch = paramTypeArray.exec(typeName);
+    if (arrayMatch) {
+        const [_, arrayItemTypeName, arraySize] = arrayMatch;
+        const entryType = getFieldType(arrayItemTypeName, types);
+        if (entryType.data_type === Enum_EthereumDataType.ARRAY) {
+            throw ERRORS.TypedError('Method_InvalidParameter', 'Nested arrays are not supported');
```
matejcik: you don't need to reproduce the failure from trezorlib :) host side can and should support encoding nested arrays. we can fix this on the firmware side independently.

matejcik (commented):
i'm OK with using this code, you seem to be generally better at javascript than I am :) so this PR is the more complete one of the two

src/js/core/methods/helpers/ethereumTypeDataConversions.js
aloisklink: I'll need to double check if EIP-712 supports that this evening. If it doesn't, I'll update the error message to something like `EIP-712 doesn't support nested arrays`.

I believe solidity (Ethereum EVM programming language) doesn't support nested arrays, and that's where EIP-712 types come from.

matejcik: The spec doesn't say either way, so we'd need to look into implementations. Metamask's eth-sig-util does not have test vectors for nested arrays, so we could guess that they don't support it.

mroz22 (changes requested):
It looks very well overall. Only if you could please: 
- add some very basic unit tests for `helpers/ethereumTypeDataConversions.js`. Not really needed to get 100% coverage. Basic tests for exported methods will be enough.
- Use Buffer instead of TextEncoder to keep it consistent with the rest of the codebase. Is that possible for you?

src/js/core/methods/helpers/ethereumTypeDataConversions.js
aloisklink: Done in 0103d17334c81e9ba7e1b7b3ed39c7b396f35ba7

Although, it does hurt slightly to use a Node.JS library instead of built-in tool in both Web/Node due to having to support Internet Explorer :sob:

src/js/core/methods/helpers/ethereumTypeDataConversions.js
aloisklink: Good catch! Fixed in c2e9887de3b6c9e51296aad2701e3cf21f35d669

src/js/core/methods/helpers/ethereumTypeDataConversions.js
aloisklink: I've removed this error and added some unit tests for type parsing in 36d1201398ffc43e507922e565c06b5add20ee27.

I couldn't find anywhere that explicitly says it's supported, but Metamask's `signTypedData_v4`'s implementation seems to support it in: https://github.com/MetaMask/eth-sig-util/blob/5f2259463990050606cd58b43e64e4bffcb715f4/src/sign-typed-data.ts#L202-L211

They use the last `[]` in the string to decode array types, e.g. `uint8[1][2]` would be an array of `uint8[1]` of length `2` (aka C-order/row-major, not Fortran-order).

aloisklink:
Description of commits I made, since they're pretty much all just called `fixup! ...`

- https://github.com/trezor/connect/pull/983/commits/0103d17334c81e9ba7e1b7b3ed39c7b396f35ba7 I replaced `TextEncoder` with `Buffer.from(x, "utf-8).toString("hex")` for Internet Explorer support.
- https://github.com/trezor/connect/pull/983/commits/bcfab256f531da3351dc53f4578a733a263482f9 Reverts https://github.com/trezor/connect/commit/34d9392debc4f27963dd972c4be45ea42b52fed3 which added support for `TextEncoder` in jest tests. We'll manually delete this commit and the original before merging, since I don't think you can use `fixup!` to delete a commit.
- https://github.com/trezor/connect/pull/983/commits/c2e9887de3b6c9e51296aad2701e3cf21f35d669 Fix misspelling in `twosComplement()`
- https://github.com/trezor/connect/pull/983/commits/be4e0765f42bc6c6b4faf9466071ca317c9d5b8e Unit tests added! It was a good idea to add them too! I wrote a `-1` instead of a `+1` when trying to calculate when a two's compliment conversion should overflow.
  - I also renamed `helpers/ethereumTypedDataConversions.js` just to `helpers/ethereumTypedData.js`.
  - I changed all `throw new Error(...` with `throw ERRORS.TypedError('Method_InvalidParameter', ...`
- https://github.com/trezor/connect/pull/983/commits/1c1fea221430667dc0d90eec8d0d42692a9220b3 I removed `bigint` and string number integration tests, since they're now covered by unit tests.
- https://github.com/trezor/connect/pull/983/commits/a7d3656d3c0a8fcf1fbdaf53b8aab2a192a4e774 Test and handle all integer overflow scenarios.
- https://github.com/trezor/connect/pull/983/commits/36d1201398ffc43e507922e565c06b5add20ee27 Remove an error message for nested arrays, as Trezor Firmware may add support for it down the line ([discussion](https://github.com/trezor/connect/pull/983#discussion_r771256372))

Just missed one more:

- https://github.com/trezor/connect/pull/983/commits/af80c5c6e35e451a156e28c10a65f9df44376581 Refactor some BigNumber.js code. Replaced a `.plus(x.negated())` with `.minus(x)` and removed the redundant arg in `.isNegative()`.

src/js/core/methods/helpers/ethereumTypeDataConversions.js
matejcik: > They use the last [] in the string to decode array types

i recall this being explicitly specified _somewhere_, but can't find it now

src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -0,0 +1,172 @@
...
+        );
+
+        // sending all the type data
+        while (response.type === 'EthereumTypedDataStructRequest') {
+            // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+            const { name: typeDefinitionName } = response.message;
+            const typeDefinition = types[typeDefinitionName];
+            if (typeDefinition === undefined) {
+                throw ERRORS.TypedError(
+                    'Method_InvalidParameter',
```
mroz22: Please change `Method_InvalidParameter` to `Runtime`.
`Method_InvalidParameter` should (in theory) refer to problem with parameters provided by implementator and should most often occur in method constructor. There are couple of more places where this requirement also applies.

src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -0,0 +1,172 @@
...
+            const dataStruckAck: EthereumTypedDataStructAck = {
+                members: typeDefinition.map(({ name, type: typeName }) => ({
+                    name,
+                    type: getFieldType(typeName, types),
+                })),
+            };
+            response = await cmd.typedCall(
+                'EthereumTypedDataStructAck',
+                // $FlowIssue typedCall problem with unions in response, TODO: accept unions
+                'EthereumTypedDataStructRequest|EthereumTypedDataValueRequest|EthereumTypedDataSignature',
```
mroz22, line 94: This is more a note for me or a question: can response here really be of type `EthereumTypedDataSignature`? I will check on this.

src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -0,0 +1,172 @@
...
+                {
+                    value: encodedData,
+                },
+            );
+        }
+
+        // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+        const signatureMessage: EthereumTypedDataSignature = response.message;
+        return {
+            address: toChecksumAddress(signatureMessage.address, network),
```
mroz22: @matejcik we don't need to checksum address in response from device anymore right? this only applies for some older firmwares and since this method is allowed only for 2.4.3 and higher we shouldn't need to do that in connect?

src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -0,0 +1,172 @@
...
+
+        this.requiredPermissions = ['read', 'write'];
+
+        const { payload } = message;
+
+        // validate incoming parameters
+        validateParams(payload, [
+            { name: 'path', obligatory: true },
+            { name: 'data', type: 'object', obligatory: true },
+            { name: 'metamaskV4Compatibility', type: 'boolean', obligatory: true },
```
mroz22: Recently we have settled that we would like to keep public interface on connect as similar to protobuf messages as possible - to be specific here, we want to use snake case for parameters that are part of protobuf so that we don't have to tranform between various object shapes. Please change `metamaskV4Compatibility` to `metamask_v4_compat`

In **future** we would also change `path` to `address_n`, but **not** in this PR just to stay consistent with other methods.

src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -0,0 +1,172 @@
...
+                // $FlowIssue typedCall problem with unions in response, TODO: accept unions
+                'EthereumTypedDataStructRequest|EthereumTypedDataValueRequest|EthereumTypedDataSignature',
+                dataStruckAck,
+            );
+        }
+
+        // sending the whole message to be signed
+        while (response.type === 'EthereumTypedDataValueRequest') {
+            // $FlowIssue disjoint union Refinements not working, TODO: check if new Flow versions fix this
+            const valueRequestMessage: EthereumTypedDataValueRequest = response.message;
```
mroz22: Total nitpick: I would rather destructure response.message directly here on order to avoid defining a new variable `valueRequestMessage`

src/js/core/methods/helpers/ethereumSignTypedData.js
```diff
@@ -0,0 +1,190 @@
...
+ * E.g. `uint16[32]` will return `{entryTypeName: 'uint16', arraySize: 32}`.
+ */
+export function parseArrayType(arrayTypeName: string): {
+    entryTypeName: string,
+    arraySize: number | null,
+} {
+    const arrayMatch = paramTypeArray.exec(arrayTypeName);
+    if (arrayMatch === null) {
+        throw ERRORS.TypedError(
+            'Method_InvalidParameter',
```
mroz22: Please change occurences `Method_InvalidParameter` to `Runtime` in this file.

mroz22 (changes requested):
Hello @aloisklink 
thanks for the latest changes you have made. They look good.

Now I believe this is the last batch of changes requests. As you can see, these are mostly nitpicks but we would like to have them resolved to keep codebase consistent.

src/js/core/methods/EthereumSignTypedData.js
aloisklink: :+1: Done in https://github.com/trezor/connect/pull/983/commits/55f3912007695f0fcb324b7ca41aac00ffcfc92e

src/js/core/methods/EthereumSignTypedData.js
aloisklink: Done in https://github.com/trezor/connect/pull/983/commits/d9bbb80f2e5bc78e37272737baf9ce856bfdaf52 along with the same changes in `src/js/core/methods/helpers/ethereumSignTypedData.js`.

src/js/core/methods/helpers/ethereumSignTypedData.js
aloisklink: Done in https://github.com/trezor/connect/pull/983/commits/d9bbb80f2e5bc78e37272737baf9ce856bfdaf52 along with the same changes in `src/js/core/methods/EthereumSignTypedData.js`.

src/js/core/methods/EthereumSignTypedData.js
aloisklink: Done in https://github.com/trezor/connect/pull/983/commits/da7872c2e91df03936e2c2c7c9ffa97798237430

My reason for adding it in was because Flow couldn't decode the types properly (it really doesn't like `await cmd.typedCall()` with union types, and flow thinks it might still be a `Promise<>`).

You didn't mention it, but I also removed the `const signatureMessage: EthereumTypedDataSignature = response.message;` in the same commit, since I was also defining a new variable `signatureMessage` instead of destructuring `response.message` directly.

src/js/core/methods/EthereumSignTypedData.js
aloisklink: My gut feeling is **maybe** for signing an empty message, e.g. something like:

```js
const fixture = {
  "path": "m/44'/60'/0'/0/0",
  "metamask_v4_compat": true,
  "data": {
      "types": {
          "EIP712Domain": [],
      },
      "domain": {},
      "primaryType: "EIP712Domain",
      "message": {},
   },
};
```

It looks like MetaMask has tests for this scenario, so although I can't see why somebody might want to sign a completely empty message, I guess it is possible: https://github.com/MetaMask/eth-sig-util/blob/73ace3309bf4b97d901fb66cd61db15eede7afe9/src/sign-typed-data.test.ts#L3258-L3275

mroz22: ah sry, I meant to put this comment on line 66. Wondering if  `EthereumSignTypedData` call might result directly into  `EthereumTypedDataSignature`

mroz22: But nevermind, this was a nitpick. I don't thing we need to change anything here.

mroz22:
Great, from my perspective this is ready to go. I would still like to wait for @matejcik to resolve his conversation and my question on whether we need to do the client side checksum (I believe we don't).

src/js/core/methods/EthereumSignTypedData.js
matejcik: I even think that this only applies to firmwares too old to do Ethereum calls, but I would have to verify.

OTOH probably no reason to not do this in on the host as well, no?

src/js/core/methods/helpers/ethereumTypeDataConversions.js
aloisklink: Ah, found it. https://github.com/ethereum/EIPs/pull/2190/files#diff-110e1894f4319a588e3e0d1ea324352f38af85215585c25029673e1ae5c6044dL140-R140

It's specified in an extension to EIP-712, which was unfortunately auto-closed after being stale for 9 months https://github.com/ethereum/EIPs/pull/2190

However, it does look like @cag and @MetaMask implemented those changes as part of `signTypedData_v4`: https://github.com/MetaMask/eth-sig-util/pull/54, and that's what most Dapps in the Ethereum eco-system uses.

src/js/core/methods/EthereumSignTypedData.js
mroz22: I think there is a reason to drop this if not needed anymore - we could get rid of another bunch of libs which is something we always like to do.

mroz22 (approved):

mroz22:
Ok, so please now squash your commits. I think this can all fit into 1 commit? Do you think it makes sense? Or is it better to keep some of the commits separated?

aloisklink:
Hi @mroz22, I merged everything into one commit https://github.com/trezor/connect/pull/983/commits/30d17ab12d445365f08c47707d9b8d5006751463, except the change in `.eslintrc` that allows variables starting with `_` to be unused, which I left in https://github.com/trezor/connect/pull/983/commits/82f58e34fdd543b1eb264cb4336ef1a0b0182661.

(I use a bunch of `[ _, group1, group2 ] = regexMatch;` for doing regex matches).

Let me know if you want me to merge them into one commit too.

aloisklink:
Is it worth me rebasing on the `develop` branch due to https://github.com/trezor/connect/pull/988?

There might be some style errors/merge conflicts in typescript files otherwise.

mroz22:
Yes, please also rebase and fix potential conflicts and I believe we are done after that.

aloisklink:
Rebased onto f0009cf412456e152dbb99cdb283f879f6c6ac84 (aka the current `develop` HEAD).

No merge conflicts, but there were a couple of TypeScript style issues that went away after a `yarn run lint:fix`.
(I've already squashed them into https://github.com/trezor/connect/pull/983/commits/7075456d45a6033b56cee7ea5373756f8b22fdf5)

PR #985, opened 2021-12-19
test: Improve OCI containers in `tests/run.sh`

aloisklink:
Two quality of life changes to running containers in `tests/run.sh`, since I ran this script multiple times for my PR #983.

#### Firstly, I added the flag `-D DOCKER_PATH` to `tests/run.sh`.

This allows running the script with `tests/run.sh -D podman`, using podman instead of docker to run containers.

`podman` has an almost identical CLI to `docker`, but it's:

  - open-source (no worry about Docker's licensing rules)
  - daemonless (only uses resources when a container is actually running, and more secure since it doesn't need root)

#### Secondly, I added the `--rm` flag to `docker run` ([see docs](https://docs.docker.com/engine/reference/run/#clean-up---rm)).

This automatically deletes the container when the container exits, so you don't have loads of unused containers left over after running integration tests a few times. If you run `docker ps --all`, you might see loads of stopped containers on your computer!

Docker keeps it by default in case you want to explore the file-system of the container/restart it later on, but I don't think that'd be common when testing `trezor-connect`.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -21,7 +21,9 @@ _Note: Running tests with custom firmware [is not currently possible](https://gi
 
 ### On Linux
 #### Prerequisites
+
 - [Docker](https://docs.docker.com/engine/install/)
+- [Podman](https://podman.io/getting-started/installation) using `tests/run.sh -D podman` (not officially supported)
 
 ### On MacOS
 _Note: As of now M1 Macs aren't supported. See [this issue](https://github.com/trezor/trezor-suite/issues/3616) for detailed information._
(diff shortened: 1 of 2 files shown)


mroz22:
Looks cool, something I need to learn about :]. Could you please also check on this PR @vdovhanych ?

vdovhanych:
It's looking nice, not sure about removing the container when it's done, but it's generally not a big deal either way. As to podman I think it's a good addition 👍

vdovhanych (approved):

PR #990, opened 2021-12-21
chore: remove ethereum checksum address

mroz22:
since min required versions to work with eth are `1.8.0` and  `2.1.0`, we can most likely remove client side address checksum.

### DIFF
--- a/src/js/device/DeviceCommands.js
+++ b/src/js/device/DeviceCommands.js
@@ -16,19 +16,12 @@ import {
     toHardened,
 } from '../utils/pathUtils';
 import { getAccountAddressN } from '../utils/accountUtils';
-import { toChecksumAddress } from '../utils/ethereumUtils';
 import { versionCompare } from '../utils/versionUtils';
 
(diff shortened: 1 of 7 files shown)


src/js/device/DeviceCommands.js
```diff
@@ -278,18 +271,15 @@ export default class DeviceCommands {
         };
     }
 
-    async ethereumGetAddress(
-        { address_n, show_display }: PROTO.EthereumGetAddress,
-        network?: EthereumNetworkInfo,
-    ) {
+    async ethereumGetAddress({ address_n, show_display }: PROTO.EthereumGetAddress) {
```
mroz22, line 274: can be removed

mroz22: ah, it is probably useful to keep it here as it does some additional work (path serialization).

mroz22:
I think this is ready to merge @szymonlesisz

szymonlesisz (approved):

PR #993, opened 2021-12-22
Refactor/method payload typed

szymonlesisz:
Resolving long term issue where `CoreMessage.payload` is not typed (any)

`AbstractMethod` requires generic type, key of API interface (key == method name) 
with this key it's now able to assign proper type to `payload` object and pass it typed to child (particular method)

`constructor` of each method is no longer required (processed by AbstractMethod) and it was replaced by `init` function (called in Core)

This refactor revealed some minor issues, mostly on type level: missing params in API but used in runtime, correct unions `xxx?: typeof undefined`

There are few runtime changes but nothing breaking

Other issues will be addressed later trezor/trezor-suite#4887


szymonlesisz:
closing in favor of https://github.com/trezor/connect/pull/996

PR #996, opened 2021-12-22
Refactor/params validator

### LINKED ISSUE #770: paramsValidator and optional types
Currently the paramsValidator helper used mostly to validate the message payloads doesn't seem to support optional types in the sense that the value itself would be null - either the key is completely missing or it is present and has the given type.

https://github.com/trezor/connect/blob/d38be2d259a2be87dbadfd2a24e476ed2cabe247/src/js/core/methods/helpers/paramsValidator.js#L22

Is that intended? Because it makes the integrations dealing with optional values kind of awkward, requiring them to delete null keys from the object before passing it to connect.

Moreover, the meaning of "obligatory: false" can be seen as kind of misleading as somebody may (imho legitimately) expect that is equivalent to the value being "optional", hence wrongly supposing a key having a null value passes just fine through the validation.

Would it make sense to add to paramsValidator an "allowNull" flag alongside "allowEmpty" to allow for a null value to be passed without breaking the existing validation rules? Or is there already some cleaner way to achieve the desired handling of null/undefined values that I'm just missing?

szymonlesisz:
❄❄❄❄ 🎄🎄🎄 ⛄⛄⛄⛄⛄  XMASS  CODE CLEANUP 🌨🌨  🎁 🎁 🎁 ❄❄❄❄

based on https://github.com/trezor/connect/pull/993

- refactored `paramsValidator` function, now it ignores undefined/null values which are not required (obligatory), fix https://github.com/trezor/connect/issues/770
- reword `paramsValidator` fields: `obligatory > required` and `amount > uint`
- protobuf types (patches) are now using shared `type UintType = string | amount`, part of trezor/trezor-suite#4887
- removed dead `debugLink` methods part of trezor/trezor-suite#4887


src/js/types/trezor/management.js
```diff
@@ -21,8 +22,14 @@ export type ChangePin = {
 
 export type FirmwareUpdateBinary = {
```
mroz22, line 23: changes seem to be missing in management.d.ts

src/js/types/misc.js
```diff
@@ -17,9 +17,11 @@ export type CipheredValue = {
 export type LoginChallenge = {
     challengeHidden: string,
     challengeVisual: string,
+    callback?: typeof undefined,
+    asyncChallenge?: typeof undefined,
```
mroz22, line 21: not part of misc.d.ts

src/js/types/misc.js
```diff
@@ -17,9 +17,11 @@ export type CipheredValue = {
...
 };
 
-export type RequestLoginAsync = { callback: () => LoginChallenge };
+export type RequestLoginAsync = { callback: () => LoginChallenge, asyncChallenge?: boolean };
```
mroz22, line 24: not part of misc.d.ts

src/js/core/methods/helpers/cardanoTokenBundle.js
```diff
@@ -12,18 +12,18 @@ export type AssetGroupWithTokens = {
 const validateTokens = (tokenAmounts: CardanoToken[]) => {
     tokenAmounts.forEach(tokenAmount => {
         validateParams(tokenAmount, [
-            { name: 'assetNameBytes', type: 'string', obligatory: true },
-            { name: 'amount', type: 'amount' },
-            { name: 'mintAmount', type: 'amount' },
+            { name: 'assetNameBytes', type: 'string', required: true },
+            { name: 'amount', type: 'uint' },
+            { name: 'mintAmount', type: 'uint', allowNegative: true },
```
mroz22, line 17: maybe distinction `int` vs `uint` would make sense. could be discussed any time later.

src/js/core/methods/GetAccountInfo.js
```diff
@@ -97,13 +96,18 @@ export default class GetAccountInfo extends AbstractMethod {
...
+
+            return {
+                ...batch,
+                address_n,
+                coinInfo,
+            };
         });
 
-        this.params = payload.bundle;
+        // this.params = payload.bundle;
```
mroz22: ? aha, this.params is already set and validated from above.

src/js/types/account.js
```diff
@@ -173,12 +173,14 @@ export type AccountInfo = {
 export type RegularOutput = {
     type?: 'external',
     address: string,
+    address_n?: typeof undefined,
```
mroz22, line 176: `address_n` account.d.ts? or maybe this is just some flow union magic right?

CHANGELOG.md
```diff
@@ -75,7 +75,7 @@
 
 ### Changed
 - Cardano: since transaction streaming has been introduced, it isn't possible to return the whole serialized transaction from the `cardanoSignTransaction` call anymore. Instead the transaction hash, transaction witnesses and auxiliary data supplement are returned and the serialized transaction needs to be assembled by the client.
-- Cardano: an obligatory `signingMode` parameter has been added to the `cardanoSignTransaction` call
+- Cardano: an required `signingMode` parameter has been added to the `cardanoSignTransaction` call
```
mroz22: nitpick typo: `a required`

src/js/core/methods/CardanoSignTransaction.js
```diff
@@ -107,31 +108,30 @@ export default class CardanoSignTransaction extends AbstractMethod {
...
-            { name: 'signingMode', type: 'number', obligatory: true },
-            { name: 'inputs', type: 'array', obligatory: true },
-            { name: 'outputs', type: 'array', obligatory: true, allowEmpty: true },
-            { name: 'fee', type: 'amount', obligatory: true },
-            { name: 'ttl', type: 'amount' },
+            { name: 'signingMode', type: 'number', required: true },
+            { name: 'inputs', type: 'array', required: true },
+            { name: 'outputs', type: 'array', required: true, allowEmpty: true },
+            { name: 'fee', type: 'uint', required: true },
+            { name: 'ttl', type: 'uint' },
```
mroz22, line 115: I can see that protobuf has this field as optional, but should it be set as required in this runtime check? I mean it used to be defined as required in flow.

mroz22 (approved):

src/js/types/trezor/management.js
szymonlesisz: all `typeof undefined` in type unions are required only by flow

src/js/types/misc.js
szymonlesisz: all `typeof undefined` in type unions are required only by flow

src/js/types/account.js
szymonlesisz: all `typeof undefined` in type unions are required only by flow

src/js/types/misc.js
szymonlesisz: good catch!

szymonlesisz: fixed

src/js/core/methods/GetAccountInfo.js
szymonlesisz: removed

PR #1003, opened 2022-01-05
Fix/electron fixes

szymonlesisz:
Prerequisites for https://github.com/trezor/trezor-suite/pull/4704

### DIFF
--- a/docs/methods/getAccountInfo.md
+++ b/docs/methods/getAccountInfo.md
@@ -64,7 +64,7 @@ TrezorConnect.getAccountInfo({
 Get info about account using public key (device is not used)
 ```javascript
 TrezorConnect.getAccountInfo({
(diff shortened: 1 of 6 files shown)


mroz22 (approved):
makes sense to me. just one question: 

I see two reasons for using webworkers in general: 
- improve performance of the main thread
- code splitting, load only those workers that are needed.

I understand that we don't really need code-splitting in node.js as everything is part of the bundle. And performance here is not a problem as well right? It is because communication with backends is quite simple task in terms of computing?

szymonlesisz:
it's still being used as a worker (TinyWorker), only difference is that now the worker loaded from the source (./lib directory) not from blockchain-link build. We need to get rid of that "build" from blockchain-link npm

PR #986, opened 2021-12-19
feat: Add EIP-712 support for T1

alisinabh:
Based on #983

This PR adds support for T1 blind signing to Trezor-Connect. It uses `@metamask/eth-sig-util` package to calculate the hashes for domain and message then sends them to T1 using `EthereumSignTypedHash` message. Since T1 will return `EthereumTypedDataSignature` no other change is needed.

This way we handle T1 and TT differences as an internal logic of `connect` and will be invisible to the user/developers.

One of the tests is failing. It is titled as `struct_list_non_v4`. I think we should somehow disable this for T1 (or completely remove it) as in a non V4 this is not supported. The `eth-sig-util` will throw the error 
```
Arrays are unimplemented in encodeData; use V4 extension
```
It might be an invalid example IMO.

Also should I move the new dependency `@metamask/eth-sig-util` to `extendedDependencies` @matejcik?

## Tasks
 - [ ] Add `domain_hash` and `message_hash` to the test fixtures.
 - [ ] Update the [trezor-common](https://github.com/trezor/trezor-common) repo and sync the submodule.
 - [ ] ~~Decide about `struct_list_non_v4` failing test.~~
 - [x] Rebase on develop after #983 is merged.
 - [ ] Re-run `scripts/protobuf-build.sh`.
 - [ ] Add `ethereumSignTypedData` to CI tests.
 - [ ] Edit [TrezorConnect.ethereumSignTypedData](methods/ethereumSignTypedData.md) instructions.
 - [ ] Mark as Ready for review.

PR #1015, opened 2022-01-13
Add t1 eip 712 support

mroz22:
based on #986


src/js/core/methods/EthereumSignTypedData.js
```diff
@@ -36,21 +42,54 @@ export default class EthereumSignTypedData extends AbstractMethod<'ethereumSignT
...
+        this.params = { ...payload, address_n: path };
     }
 
     async run() {
         const cmd = this.device.getCommands();
-        const { path: address_n, data, metamask_v4_compat } = this.params;
+        const { address_n, metamask_v4_compat } = this.params;
+
+        if (this.device.features.model === '1') {
+            if (!this.params.metamask_v4_compat) {
```
matejcik: This should not be here, only in the plugin. T1 will happily sign hashes generated as-if with `metamask_v4_compat=false` -- it's the plugin that is unable to _generate_ this data.

matejcik (commented):
general note: we have https://github.com/trezor/trezor-firmware/pull/2036 in the pipeline, which makes `message_hash` optional in case `primary_type == "EIP712Domain"` (in this case `message` should be supplied but it can be `{}`

once that is in, Connect should support that usecase, namely something like:
```javascript
if primaryType == "EIP712Domain" {
  message_hash = null
}
```
i'll leave it to you if we want to future-proof the Connect code right now, or if we want to release as-is and push an update later

src/js/core/methods/EthereumSignTypedData.js
mroz22: thanks, removed.

szymonlesisz (approved):

PR #1035, opened 2022-01-30
Remove AOPP support

prusnak:
Not sure I got everything correctly. Please review @szymonlesisz @mroz22


mroz22 (approved):
I think that should be it.

PR #1033, opened 2022-01-27
Support domain-only EthereumSignTypedData and other improvements

aloisklink:
Add documentation on using `EthereumSignTypedData` on T1 and improved types. These were missing from #1015.

### Minor breaking change

`EthereumSignTypedData` now requires passing Trezor T variables. This seems makes sense from my side, since everybody using this function should pass Trezor T variables anyway, and since this function was only recently released, I doubt this change would affect any users of this library.

Plus, it makes all the code (especially types!) simpler.

Trezor Model 1 parameters are still optional, as blind signing requires using an extra dependency on an external library, and is less secure.

### Support for `primaryType` = `EIP712Domain`

~I tried adding docs/tests/types for https://github.com/trezor/trezor-firmware/pull/2036 in this PR too, but updating the `submodules/trezor-common` submodule just causes a bunch of other Protobuf type errors to occur, so I'll do that in another separate PR in the future.~

I've added support for https://github.com/trezor/trezor-firmware/pull/2036 in this PR too, since they both edit similar files.

I've tested locally using:

- `./tests/run.sh -D 'podman' -i 'ethereumSignTypedData'` Success now #1042 been merged ~Fails~
- `./tests/run.sh -D 'podman' -i 'ethereumSignTypedData' -f '2-master'` Success
- `./tests/run.sh -D 'podman' -i 'ethereumSignTypedData' -f '1-master'` Success

PR #928, opened 2021-11-01
single messages.json 

mroz22:
# Task: 

use only one messages.json file

# Versions 

|    | connect                 | trezor-common | more than 2 years                 | max fw incl |                   
|----|-------------------------|---------------|-----------------------------------|------------|
| v6 | b3af21e ~ July 27, 2018 | https://github.com/trezor/trezor-common/tree/4c77015ed9c964603e193b31fb583b8adf31bea8             | Yes              | `1.7.3` `2.0.10` |
| v7 | d5b02dd ~ Aug 7, 2019   | https://github.com/trezor/trezor-common/tree/868b7db8826780f253f5068ad483406e633594a4               | Yes             | `1.8.1` `2.1.1` |
| v8 | 50432fa ~ Feb 18, 2020  | https://github.com/trezor/trezor-common/tree/669fc83eb0fdc8c5aecbfc5a4ae328d1ef66e786              | No               | `1.8.9` `2.2.9` |

# Firmwares 

According to the deprecation strategy we are bound to support 2 yeas old firmwares and all bootloaders. 

So the first firmwares we may stop supporting are 1.8.3 and 2.1.5 https://github.com/trezor/webwallet-data/commit/69fd179a0d8c765c179acb7f719eac98554c3d8d

```
   {
        "required": false,
        "version": [1, 8, 3],
        "bootloader_version": [1, 8, 0],
        "min_bridge_version": [2, 0, 25],
        "min_firmware_version": [1, 6, 2],
        "min_bootloader_version": [1, 5, 0],
        "url": "data/firmware/1/trezor-1.8.3.bin",
        "url_bitcoinonly": "data/firmware/1/trezor-1.8.3-bitcoinonly.bin",
        "fingerprint": "496aecfab867504b2283a9f057a0b2fd9d17970a22c81f6ad74232e7b914ce68",
        "fingerprint_bitcoinonly": "13d6089cb935f453eaddbfe193e0ab37924a7aa66f684355a4fe5c660c18247a",
        "changelog": "* Small code improvements"
    },

    {
        "required": false,
        "version": [2, 1, 5],
        "min_bridge_version": [2, 0, 7],
        "min_firmware_version": [2, 1, 0],
        "min_bootloader_version": [2, 0, 0],
        "url": "data/firmware/2/trezor-2.1.5.bin",
        "url_bitcoinonly": "data/firmware/2/trezor-2.1.5-bitcoinonly.bin",
[...]

PR #1042, opened 2022-02-04
Fix `getFirmwareRange()` for multiple matching firmware ranges

aloisklink:
Fixes a bug in the `getFirmwareRange()` function.

If there are multiple matching rules in [`src/data/config.json`](https://github.com/aloisklink/connect/blob/d0ee310534d15e9be2f436a1e35746ff6b29349a/src/data/config.json#L97-L101), `getFirmwareRange()` only uses the first rule.

Because there was a rule called `coin: ["eth", ` at the beginning of this file, all of the other Ethereum rules were being ignored. This meant that there was no `ui-device_firmware_unsupported` error thrown when a firmware didn't support EIP1559 or `ethereumSignTypedData`.

This PR changes `getFirmwareRange()` so that it goes through all matching rules in `src/data/config.json`.

### Test Results

- unit tests pass
- emulator tests:
  - **all pass** when running `./tests/run.sh -D 'podman' -f '2-master'`
  - **failures** on ethereumTypedData when running `./tests/run.sh -D 'podman'`
    - I'll fix in #1033
  - **failures** when running `./tests/run.sh -D 'podman' -f '1-master'`
    - I added `legacyResults` to the tests that were failing since the T1 firmware was unsupported in https://github.com/trezor/connect/commit/d0ee310534d15e9be2f436a1e35746ff6b29349a.
    - fails on ethereumTypedData (which I'll fix in #1033),
    - finally, a bunch of functions consistently fail in `signTransaction` (e.g. `Testnet (P2PKH)`) with errors:
      - `Device_CallInProgress`
      - `Async callback was not invoked within the 20000 ms timeout specified by jest.setTimeout.Error`

Edit: :facepalm: Forgot to run `yarn lint`, I'll force-push the changes since I don't think anybody would have reviewed the changes in the last few minutes.

### DIFF
--- a/src/js/core/methods/helpers/__fixtures__/paramsValidator.js
+++ b/src/js/core/methods/helpers/__fixtures__/paramsValidator.js
@@ -233,17 +233,18 @@ export const getFirmwareRange = [
         result: { '1': { min: '1.10.0', max: '0' }, '2': { min: '2.4.0', max: '0' } },
     },
     {
-        description: 'range from config.json (by coinType)',
+        description: 'range from config.json (by coinType and coin as string)',
         config: {
             supportedFirmware: [
                 // this one is ignored, different excludedMethod
-                { coinType: 'bitcoin', methods: ['showAddress'], min: ['1.11.0', '2.5.0'] },
-                { coinType: 'bitcoin', min: ['1.10.0', '2.4.0'] },
-                { coin: 'btc', min: ['1.11.0', '2.5.0'] },
+                { coinType: 'bitcoin', methods: ['showAddress'], min: ['1.12.0', '2.6.0'] },
+                // should merge both of these ranges together, since they both match
+                { coinType: 'bitcoin', min: ['1.10.0', '2.5.0'] },
+                { coin: 'btc', min: ['1.11.0', '2.4.0'] },
             ],
         },
         params: ['signTransaction', DEFAULT_COIN_INFO, DEFAULT_RANGE],
-        result: { '1': { min: '1.10.0', max: '0' }, '2': { min: '2.4.0', max: '0' } },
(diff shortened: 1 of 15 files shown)


src/js/core/methods/helpers/__fixtures__/paramsValidator.js
```diff
@@ -362,6 +363,6 @@ export const getFirmwareRange = [
             { support: { trezor1: '1.6.2', trezor2: '2.1.0' }, shortcut: 'eth', type: 'ethereum' },
             DEFAULT_RANGE,
         ],
-        result: { '1': { min: '1.8.0', max: '0' }, '2': { min: '2.1.0', max: '0' } },
+        result: { '1': { min: '1.10.4', max: '0' }, '2': { min: '2.4.2', max: '0' } },
```
aloisklink, line 366: This reverts the change made in https://github.com/trezor/connect/pull/928/commits/9fecde426e45dfb702f000902ce6544d5844f484

mroz22 (approved):
looks great
we also have gitlab CI where we are running additional tests for older firmwares and it seems they are passing with this branch https://gitlab.com/satoshilabs/trezor/connect/-/pipelines/465831004

FrederikBolding:
Stumbled upon this bug when testing the MetaMask integration that @aloisklink has worked on - Any news on when this fix is released? @mroz22
With this bug users aren't gracefully migrated to the new firmware but will be seeing errors instead.

mroz22:
Hi, we are preparing 8.2.7 release now. If everything goes well, it should be out tomorrow around 12.00 CET

FrederikBolding:
Sounds good! Thanks for the update @mroz22 😄

PR #1043, opened 2022-02-04
chore(nix): update to match trezor-suite

### LINKED ISSUE #732: Nix shell simplification broke yarn flow
The commit 0025fff2 introduced some changes and it fails for `yarn flow`. I have reverted this commit in develop 3f1336bf0b08fd827d0dc039c712f777589a62c3 because it was blocking @szymonlesisz.

See this [job](https://gitlab.com/satoshilabs/trezor/connect/-/jobs/916558526). Reproduced on my NixOS.

**How to reproduce**

1. `nix-shell`
2. `yarn`
3. `yarn flow`

**Notes**
If we now install `flow` natively I guess we need to link so it is taken from nix store and not bin modules? I remembered we did smth like that in Suite but there are no symlinks [there](https://github.com/trezor/trezor-suite/blob/develop/shell.nix) now.

### LINKED ISSUE #4891: nix-shell simplification in Connect broke yarn flow
The commit 0025fff2 introduced some changes and it fails for `yarn flow`. I have reverted this commit in develop 3f1336bf0b08fd827d0dc039c712f777589a62c3 because it was blocking @szymonlesisz.

See this [job](https://gitlab.com/satoshilabs/trezor/connect/-/jobs/916558526). Reproduced on my NixOS.

**How to reproduce**

1. `nix-shell`
2. `yarn`
3. `yarn flow`

**Notes**
If we now install `flow` natively I guess we need to link it from the nix store and not bin modules right? I remembered we did smth like that in Suite but there are no symlinks [there](https://github.com/trezor/trezor-suite/blob/develop/shell.nix) now. I wasn't 100% how to fix this so I have figured it is more effective to ask you :).

tsusanka:
I have tested build and eslint and it seemed to work. In case @szymonlesisz is using it he should double-check.

Closes trezor/trezor-suite#4891.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,8 +1,8 @@
-# the last successful build of nixos-20.09 (stable) as of 2020-10-11
+# the last successful build of nixpkgs-unstable as of 2021-11-16 compatible to trezor-suite
 with import
   (builtins.fetchTarball {
-    url = "https://github.com/NixOS/nixpkgs/archive/0b8799ecaaf0dc6b4c11583a3c96ca5b40fcfdfb.tar.gz";
-    sha256 = "11m4aig6cv0zi3gbq2xn9by29cfvnsxgzf9qsvz67qr0yq29ryyz";
+    url = "https://github.com/NixOS/nixpkgs/archive/5cb226a06c49f7a2d02863d0b5786a310599df6b.tar.gz";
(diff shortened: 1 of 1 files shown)


mroz22:
let's merge and see.

mroz22 (approved):

szymonlesisz (approved):

PR #1045, opened 2022-02-09
Fix/custom backend onion

szymonlesisz:


src/js/backend/BlockchainLink.js
```diff
@@ -331,9 +345,10 @@ export const initBlockchain = async (
             postMessage,
             debug: DataManager.getSettings('debug'),
             proxy: DataManager.getSettings('proxy'),
-            onionDomains: DataManager.getSettings('useOnionLinks')
-                ? DataManager.getConfig().onionDomains
-                : undefined,
+            onionDomains:
+                DataManager.getSettings('useOnionLinks') && !customBackends[coinInfo.shortcut]
```
marekrjpolak, line 352: Was it necessary to remove the .onion translation for custom backends? I don't think it's a problem either to have it or not to have it there; I'm just curious.

szymonlesisz: not a problem, it has no effect on suite.  I think we talked about it earlier that it may be confusing that you explicitly requested some custom address and get connected to different one

marekrjpolak (approved):

PR #1046, opened 2022-02-10
fix: Cardano testnet protocol magic

vladimirvolek:


mroz22 (approved):

mroz22 (changes requested):
ah, need to solve test results

PR #1050, opened 2022-02-13
Add CI tests for Ethereum plugin

aloisklink:
Adds CI tests for the Ethereum plugin (as discussed in https://github.com/trezor/connect/pull/1033#issuecomment-1033421429)

I've added the test to both GitHub Actions and GitLab CI. However, I've never used GitLab CI before, so I'm not 100% sure if this would work.

### DIFF
--- a/src/js/plugins/ethereum/package.json
+++ b/src/js/plugins/ethereum/package.json
@@ -1,6 +1,7 @@
 {
     "name": "trezor-connect",
     "version": "1.0.0",
+    "private": "true",
     "author": "Trezor <info@trezor.io>",
     "homepage": "https://github.com/trezor/connect",
     "description": "Trezor 3rd party plugins",
--- a/src/js/plugins/ethereum/typedData.test.js
+++ b/src/js/plugins/ethereum/typedData.test.js
@@ -8,26 +8,28 @@ function messageToHex(string) {
 }
 
 describe('typedData', () => {
-    commonFixtures.tests
-        .filter(test => test.parameters.metamask_v4_compat)
-        .forEach(test => {
-            it('typedData to message_hash and domain_separator_hash', () => {
-                const transformed = typedData(
-                    test.parameters.data,
-                    test.parameters.metamask_v4_compat,
-                );
-                const { domain_separator_hash, message_hash } = transformed;
-
-                expect(messageToHex(domain_separator_hash)).toEqual(
-                    messageToHex(test.parameters.domain_separator_hash),
-                );
-                if (message_hash) {
-                    expect(messageToHex(message_hash)).toEqual(
-                        messageToHex(test.parameters.message_hash),
+    describe('should create message_hash and domain_separator_hash from typedData', () => {
+        commonFixtures.tests
+            .filter(test => test.parameters.metamask_v4_compat)
+            .forEach(test => {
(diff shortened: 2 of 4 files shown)


src/js/plugins/ethereum/package.json
```diff
@@ -1,6 +1,7 @@
 {
     "name": "trezor-connect",
     "version": "1.0.0",
+    "private": "true",
```
aloisklink, line 4: I've set `"private": "true"` here so that nobody at @Trezor accidentally publishes this plugin to NPM, see https://docs.npmjs.com/cli/v8/configuring-npm/package-json#private
and https://yarnpkg.com/configuration/manifest#private

(the package name currently conflicts with `trezor-connect`, even if you did want to publish it)

src/js/plugins/ethereum/typedData.test.js
```diff
@@ -8,26 +8,28 @@ function messageToHex(string) {
...
-                const transformed = typedData(
-                    test.parameters.data,
-                    test.parameters.metamask_v4_compat,
-                );
-                const { domain_separator_hash, message_hash } = transformed;
+    describe('should create message_hash and domain_separator_hash from typedData', () => {
+        commonFixtures.tests
+            .filter(test => test.parameters.metamask_v4_compat)
+            .forEach(test => {
+                it(test.name, () => {
```
aloisklink, line 15: The main thing I changed here was to add another `describe()` call, and in the `it()` call use the fixture `test.name` as the test name. It's just to make the `yarn test:unit` results look prettier :)

mroz22:
Hi, I ran your PR in gitlab CI (which is not triggered automatically for 3rd party)
- there was some minor typo in gitlab yml syntax which I fixed in https://github.com/trezor/connect/commit/541298ab30c26204ddcf1c1c9d964b7a8afb4fe6
- now it exits with some error "nix-shell not found" and I am not sure why, at the first glance setup looks ok. So I am sharing it here and maybe you will have some idea https://gitlab.com/satoshilabs/trezor/connect/-/jobs/2099096538

But all good, we don't need to run it in gitlab that much. It actually makes best sense in github (which is triggered for 3rd party) as these plugins are develop mostly by 3rd party.

aloisklink:
Hi @mroz22,

Thanks for trying to fix the GitLab issues!

It looks like https://gitlab.com/satoshilabs/trezor/connect/-/jobs/2099096538 was just running on a generic `busybox` Docker image (probably the default one), which didn't have `nix-shell` installed.

```logs
Preparing the "docker" executor
Using Docker executor with image busybox ...
```

I've added `image: registry.gitlab.com/satoshilabs/trezor/trezor-user-env/trezor-user-env:latest`, so it should hopefully now run on your Trezor Env Docker image, which contains `nix-shell`.
https://github.com/trezor/connect/blob/118e6b3bbbdca64d4070b9b8449a8781ff2a0d19/.gitlab-ci.yml#L210-L212

If it still doesn't work (and since you said GitLab CI isn't as important as GitHub CI), I `git rebase`-ed to split up the GitHub and GitLab CI changes into two different commits, so it's easy to just delete the latest commit.

mroz22:
Hello, these tests are already part of trezor-suite monorepo so we can close it here. Thanks for contribution!

PR #1053, opened 2022-02-21
chore: update coins.json

mroz22:


matejkriz (approved):
verified by these steps: https://github.com/trezor/connect/blob/develop/docs/supported-coins.md#update-and-maintenance-in-trezor-connect

PR #1074, opened 2022-03-17
docs: Trezor Model 1 also supports EIP-1559

sime:


mroz22 (approved):
yes, this seems correct according to what I can see in config.json 
```
{
            "capabilities": ["eip1559"],
            "min": ["1.10.4", "2.4.2"],
            "comment": ["new eth transaction pricing mechanism (EIP1559) since 1.10.4/2.4.2"]
        },
```

PR #1075, opened 2022-03-17
fix(ci): fix tenv launch to reflect latest changes

vdovhanych:
this should fix the test running in the ci after we changed the native script for launching tenv.


vdovhanych:
@mroz22 I fixed it so we can merge I think.

mroz22 (approved):

PR #1079, opened 2022-03-23
docs: References to Tx input/output types

sime:
The links to `signTransaction` are dated.

I am uncertain if linking directly to a commit is the best way to 'solve' this. Thoughts?


mroz22:
Linking to a concrete commit will  probably get broken soon once we migrate files into trezor-suite. Also linking a file that is subject to frequent changes does not feel like a good practice (I know we are doing it already).

sime:
Perhaps some code generated documentation of the types would be safer. Let's leave it for post merge.

